PR #5, opened 2021-05-17
Make metadata labels optional

ereslibre:
If policy labels are not provided in the policy metadata, assume no
labels and default to None.


flavio (approved):
Looks good to me

PR #6, opened 2021-05-18
Add policy metadata retrieval

ereslibre:


flavio:
This code is coming from kwctl, right? Are you moving it there because it fits better or because you plan to use it inside of policy-server?

I didn't know where to put this code in the beginning, but I decided to leave that into kwctl as a starting point, because I didn't plan to use it inside of policy-server.

ereslibre:
Yes, did minor changes on the code, but is basically the same thing. Metadata of a policy is something that sounds to me too generic for being on kwctl and felt a bit weird on a CLI tool. If you want to wait until we have a second consumer to move it, I'm good. I was thinking that I could add some things already: the policy-server annotates the AdmissionPolicy object on the status with: author, SHA256 of the module, license... Was not going to do that right now, but I wanted to start giving more exposure to policy details to users -- the policies that are really running in their policy server.

flavio:
Ok, let's park it for a while. We still have quite some moving parts around this metadata story.

Don't close the PR yet

ereslibre:
Merging as discussed.

PR #8, opened 2021-05-29
Policy metadata introduce `annotatedBy`

flavio:
Introduce the optional field 'annotatedBy', this can be used to track  which program added the metadata to the final Wasm file.
    
This can be used to debug issues with policies.


ereslibre (approved):
Thank you @flavio!

PR #10, opened 2021-06-08
Update wasmtime-provider

flavio:
Update to a version of wasmtime-provider that uses a more recent version of wasmtime. This is required to fix a vulnerability issue.


ereslibre (changes requested):
Libraries should not have a `Cargo.lock` file

ereslibre (approved):

PR #11, opened 2021-06-15
refactor: context awareness

ereslibre:
The crate library will not start a refresh loop, and will let up to
the caller to refresh the context aware structure when it wants.

Related: https://github.com/kubewarden/kwctl/issues/40


flavio (approved):
I think this is a step in the right direction. To me it still feels strange to have this global defined inside of a library.

I think we can revisit the whole design once we have a better idea about how we're going to make everything more generic (as in, allow user to retrieve other types of resources).

PR #13, opened 2021-07-01
feat: implement policy logging

ereslibre:
Implement a waPC `kubewarden.tracing.log` host function.

Co-authored-by: Flavio Castelli <fcastelli@suse.com>


ereslibre:
After this is merged, I will tag a new version of this crate and bump `policy-server` and `kwctl`.

ereslibre:
@flavio: requesting review for a first pass. I want to add some unit tests over here.

ereslibre:
I'm getting `fatal runtime error: stack overflow` on `cargo test`, but I don't understand why yet. This could be some issue with the closures as arguments. Looking into it.

flavio:
I've added some comments to the code and squashed everything into your commit.

Everything looks good :clap:

flavio (approved):

ereslibre:
Thank you!

@flavio: minor comment, the `Default` impl for `Policy` is only for the unit tests :), so we can mimic a policy load. I say so because the comments there can lead to think otherwise.

Issue #15, opened 2021-07-12
policy-evaluator: evaluate a Rego policy converted to Wasm


flavio:
* Turn https://github.com/flavio/burrego into a crate
* Allow the creation of a policy evaluator that can evaluate Rego code

Part of https://github.com/kubewarden/policy-evaluator/issues/14

ereslibre:
https://github.com/wapc/wasmtime-provider/pull/10 is part of this work. The `burrego` crate was depending on `wasmtime` 0.28.0 and since we depend on that crate too but on version 0.24.0 -- and given [it links to the `system](https://github.com/bytecodealliance/wasmtime/blob/ebbe3997251add0278cc2de9c18dc43a1895f580/crates/fiber/Cargo.toml#L10-L14)`, only one crate version can exist in the dependency graph. Thus, as part of this work, [we had to bump the `wasmtime-provider` of the `wapc` project to 0.28.0 too](https://github.com/bytecodealliance/wasmtime/blob/ebbe3997251add0278cc2de9c18dc43a1895f580/crates/fiber/Cargo.toml#L10-L14), and so the dependency graph is solved for this project.

PR #17, opened 2021-07-20
Integrate burrego inside the policy-evaluator

ereslibre:
Initial integration. Many bits and pieces should be polished.

Related: https://github.com/kubewarden/policy-evaluator/issues/15


crates/burrego/.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
+name: burrego release
```
flavio, line 1: I think we can remove this GitHub action

crates/burrego/.github/workflows/tests.yml
```diff
@@ -0,0 +1,77 @@
+on: [push, pull_request]
```
flavio, line 1: I wonder if this GitHub action is ran, that would be nice

crates/burrego/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.40"
+clap = "2.33.3"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0.64"
+gtmpl = { git = "https://github.com/flavio/gtmpl-rust.git", branch = "printf-handle-array-and-map" }
```
flavio, line 14: I moved this fork under the `kubewarden` org

crates/burrego/src/opa/evaluation_result.rs
```diff
@@ -0,0 +1,6 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize)]
+pub struct EvaluationResult {
+    pub result: bool,
+}
```
flavio, line 6: I think this is limiting. Violations can return also explanation messages about why something has been rejected

crates/burrego/src/opa/wasm.rs
```diff
@@ -0,0 +1,440 @@
...
+            move |mut caller: Caller<'_, Option<StackHelper>>, addr: i32| {
+                let stack_helper = caller.data().unwrap();
+                let msg = stack_helper
+                    .pull_json(caller.as_context_mut(), &memory, addr)
+                    .unwrap();
+                println!("OPA abort with message: {:?}", msg);
+                std::process::exit(1);
+            },
+        );
+        linker.define("env", "opa_abort", opa_abort)?;
```
flavio, line 127: Unrelated with the PR, we will have to handle abort in different ways.

crates/burrego/src/opa/wasm.rs
```diff
@@ -0,0 +1,440 @@
...
+        //Called to emit a message from the policy evaluation.
+        //The addr refers to a null-terminated string in the shared memory buffer.
+        let opa_println = Func::wrap(
+            &mut store,
+            move |mut caller: Caller<'_, Option<StackHelper>>, addr: i32| {
+                let stack_helper = caller.data().unwrap();
+                let msg = stack_helper
+                    .pull_json(caller.as_context_mut(), &memory, addr)
+                    .unwrap();
+                println!("Message coming from the policy: {:?}", msg);
```
flavio, line 139: Same here, we will probably have to build a callback mechanism. The consumer will provide a function/closure to use to handle `abort` and `println`

src/policy_evaluator.rs
```diff
@@ -110,26 +128,51 @@ impl PolicyEvaluator {
...
+                    |_| Ok(0),
+                    Policy::new,
+                    policy_execution_mode,
+                )?;
+                let evaluator = burrego::opa::wasm::Evaluator::new(id, &policy_contents)?;
+                let policy_runtime = Runtime::Burrego(BurregoEvaluator {
+                    evaluator,
+                    entrypoint_id: 0, // TODO: let kwctl/policy-server populate this
+                    input: serde_json::from_str("{}")?, // TODO: let kwctl/policy-server populate this
+                    data: serde_json::from_str("{}")?, // TODO: let kwctl/policy-server populate this
```
flavio, line 168: nitpicking, I think it's nicer to use `json!({})`

flavio (changes requested):
Thanks for the hard work! I've left some notes here and there.

I didn't elaborate on the `TODO`, but we will have to talk about them in a different way.

We also have to discuss about how to handle this work, I wouldn't merge that into `main` yet. I would rather do PRs against a dedicated branch and then, when everything is good, do a PR from that branch into `main`. What do you think?

crates/burrego/.github/workflows/tests.yml
ereslibre: Let me check. If not, I can try some tricks like a symlink. I will report back.

crates/burrego/src/opa/evaluation_result.rs
ereslibre: Yes, the main reason I didn't enter in detail here is: do we want to have this structure on the consumable crate, or our API is meant  to return `serde_json`? I prefer this approach, and I think I can expand this struct to the documented fields -- have to check where it's documented.

crates/burrego/.github/workflows/tests.yml
ereslibre: A symlink does not work. We will have to have a targeted workflow for it in the root folder.

ereslibre: Opened https://github.com/kubewarden/policy-evaluator/issues/18

ereslibre:
> We also have to discuss about how to handle this work, I wouldn't merge that into main yet. I would rather do PRs against a dedicated branch and then, when everything is good, do a PR from that branch into main. What do you think?

I think this is the best approach. We can discuss and keep improving it. We only have to be careful if it's more than one person touching the code and force pushing to a given shared branch.

ereslibre:
Moving this PR base branch to the `policy-evaluator` repo inside the `kubewarden` org.

PR #22, opened 2021-07-22
Integrate burrego inside the policy-evaluator

ereslibre:
Initial integration. Many bits and pieces should be polished.

Related: https://github.com/kubewarden/policy-evaluator/issues/15


src/policy_evaluator.rs
```diff
@@ -186,32 +232,77 @@ impl PolicyEvaluator {
...
-                pol_val_resp
-                    .and_then(|pol_val_resp| {
-                        ValidationResponse::from_policy_validation_response(
-                            uid.to_string(),
-                            &req_obj,
-                            &pol_val_resp,
-                        )
-                    })
-                    .unwrap_or_else(|e| {
+        match &mut self.runtime {
```
flavio, line 235: I don't like to have this `switch` statement that goes over all the possible runtimes, to then execute appropriate code.

Can't we define a trait, let's say `validate`, and then have the waPC and the Burrego runtimes implement that? This would dry this code quite a bit

src/policy_evaluator.rs
```diff
@@ -186,32 +232,77 @@ impl PolicyEvaluator {
...
+                    Ok(evaluation_result) => {
+                        let evaluation_result = serde_json::from_value(
+                            evaluation_result
+                                .get(0)
+                                .unwrap()
+                                .get("result")
+                                .unwrap()
+                                .get("response")
+                                .unwrap()
+                                .clone(),
```
flavio, line 282: This looks hard to understand. Can we do something better? Adding some documentation explaining what we are getting back as a json object would help too.

Now that I recall, is this some temporary code? I'm referring to the conversation we had around how to integrate vanilla OPA responses. I guess this code will be replaced by something that tries to unmarshall the json object into and AdmissionReview object. Am I right?

src/policy_evaluator.rs
```diff
@@ -228,40 +319,53 @@ impl PolicyEvaluator {
...
-        {
-            Ok(res) => {
-                let vr: Result<SettingsValidationResponse> = serde_json::from_slice(&res)
-                    .map_err(|e| anyhow!("cannot convert response: {:?}", e));
-                vr.unwrap_or_else(|e| SettingsValidationResponse {
-                    valid: false,
-                    message: Some(format!("error: {:?}", e)),
-                })
+        match &self.runtime {
+            Runtime::Wapc(wapc_host) => {
```
flavio, line 323: Same here. I would get rid of this `switch` by defining a common trait. The waPC runtime would perform the usual validation of the settings, the Burrego one would always return `true`.

src/policy_evaluator.rs
```diff
@@ -228,40 +319,53 @@ impl PolicyEvaluator {
...
-                anyhow!(
-                    "Cannot create ProtocolVersion object from '{:?}': {:?}",
-                    res,
-                    e
-                )
-            }),
-            Err(err) => Err(anyhow!(
-                "Cannot invoke 'protocol_version' waPC function: {:?}",
-                err
+        match &self.runtime {
```
flavio, line 353: Same here, I would cleanup with a trait

src/validation_response.rs
```diff
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Serialize, Deserialize, Debug, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct ValidationResponse {
-    pub uid: String,
+    pub uid: Option<String>,
```
flavio, line 8: Why is this now optional, is that because Rego policies might not provide it back?

flavio (changes requested):
Overall great, I just left some questions and proposed some changes.

We can even merge it as it is and then do the changes I propose via dedicated PRs (I'm referring to the usage of traits)

src/policy_evaluator.rs
ereslibre: Makes a lot of sense. Going to add it.

src/policy_evaluator.rs
ereslibre: Yes, this is exactly what I was referring to. This is going to change.

src/validation_response.rs
ereslibre: Yes. This is the case, to unmarshal into `ValidationResponse` directly what's coming from the policy. We could split types and perform conversion between them, but I think is good enough. Many other things here are Optional already.

ereslibre:
> We can even merge it as it is and then do the changes I propose via dedicated PRs (I'm referring to the usage of traits)

Let me do a minor update today and we merge. Then we do the trait suggestions in following PR's.

src/policy_evaluator.rs
flavio: Great!

flavio:
> > We can even merge it as it is and then do the changes I propose via dedicated PRs (I'm referring to the usage of traits)
> 
> Let me do a minor update today and we merge. Then we do the trait suggestions in following PR's.

@ereslibre sounds good to me. But... let's also create dedicated issues for the different things we have to tackle with these upcoming PRs.

I'll start to create these issue, but we have to make sure I don't forget anything.

flavio:
I've created these issues: https://github.com/kubewarden/policy-evaluator/labels/rego

src/policy_evaluator.rs
ereslibre: Looking at the code I'm dubious now. Is the same type: the enum Runtime. Thus, adding a trait here is not going to help: is not a case of different types with same behavior, but a single type with different behaviors based on the enum value.

ereslibre: Resolving. https://github.com/kubewarden/policy-evaluator/issues/24 is filed and we can continue over there.

src/policy_evaluator.rs
ereslibre: Resolving. https://github.com/kubewarden/policy-evaluator/issues/24 is filed and we can continue over there.

ereslibre:
I consider this done. We can follow up with the subjects that we have still pending. Let me know if you find anything blocking.

ereslibre:
> I've created these issues: https://github.com/kubewarden/policy-evaluator/labels/rego

Thanks a lot for this.

src/validation_response.rs
flavio: I'm convinced about making `uid` optional. This is something that must always be set

flavio (changes requested):
LGTM, I just have some doubts about making the response `uid` optional

flavio (approved):

PR #26, opened 2021-08-11
Add policy execution mode to metadata

flavio:
Enrich the policy metadata with a new field that explains which execution mode (kubewarden, opa, gatekeeper) must be used to evaluate the policy.
    
By default, when nothing is specified, we fall back to the kubewarden execution mode.

Note well, I reused the `PolicyExecutionMode` enum to identify the type of execution mode to be used. That forced me to implement a custom serializer and deserializer for the enum. That introduces some extra code (especially the deserialization bit), but that's how serde wants things to be done.

### DIFF
--- a/src/policy_evaluator.rs
+++ b/src/policy_evaluator.rs
@@ -16,74 +16,14 @@ use crate::cluster_context::ClusterContext;
 use crate::policy::Policy;
 use crate::validation_response::ValidationResponse;
 
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, serde::Deserialize, serde::Serialize, Debug)]
 pub enum PolicyExecutionMode {
+    #[serde(rename = "kubewarden-wapc")]
     KubewardenWapc,
+    #[serde(rename = "opa")]
     Opa,
+    #[serde(rename = "gatekeeper")]
     OpaGatekeeper,
-}
-
-impl serde::ser::Serialize for PolicyExecutionMode {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::ser::Serializer,
-    {
-        let mode = match self {
-            PolicyExecutionMode::KubewardenWapc => "kubewarden",
-            PolicyExecutionMode::Opa => "opa",
-            PolicyExecutionMode::OpaGatekeeper => "gatekeeper",
-        };
-        serializer.serialize_str(mode)
-    }
-}
-
-struct StringVisitor;
-
-impl<'de> serde::de::Visitor<'de> for StringVisitor {
-    type Value = String;
-
-    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-        write!(formatter, "a string")
-    }
-
-    fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
-    where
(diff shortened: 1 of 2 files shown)


src/policy_evaluator.rs
```diff
@@ -16,12 +16,76 @@ use crate::cluster_context::ClusterContext;
...
     OpaGatekeeper,
 }
 
+impl serde::ser::Serialize for PolicyExecutionMode {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        let mode = match self {
+            PolicyExecutionMode::KubewardenWapc => "kubewarden",
```
ereslibre, line 32: We could use `"kubewarden-wapc"` and be future proof.

src/policy_evaluator.rs
```diff
@@ -16,12 +16,76 @@ use crate::cluster_context::ClusterContext;
...
+        write!(formatter, "a string")
+    }
+
+    fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        Ok(s.to_owned())
+    }
+}
```
ereslibre, line 55: Do we need this pattern?

src/policy_evaluator.rs
```diff
@@ -16,12 +16,76 @@ use crate::cluster_context::ClusterContext;
...
+            "kubewarden" => Ok(PolicyExecutionMode::KubewardenWapc),
+            "opa" => Ok(PolicyExecutionMode::Opa),
+            "gatekeeper" => Ok(PolicyExecutionMode::OpaGatekeeper),
+            _ => Err(serde::de::Error::custom(format!(
+                "{} is not a valid execution mode",
+                mode
+            ))),
+        }
+    }
+}
```
ereslibre, line 73: Maybe I'm missing something, but do we need to implement this? Cannot we derive it?

src/policy_evaluator.rs
```diff
@@ -16,12 +16,76 @@ use crate::cluster_context::ClusterContext;
...
+            PolicyExecutionMode::KubewardenWapc => "kubewarden",
+            PolicyExecutionMode::Opa => "opa",
+            PolicyExecutionMode::OpaGatekeeper => "gatekeeper",
+        };
+
+        f.debug_struct("PolicyExecutionMode")
+            .field("mode", &mode)
+            .finish()
+    }
+}
```
ereslibre, line 87: Same here, cannot we derive this one?

ereslibre (commented):
Thank you, some comments.

src/policy_evaluator.rs
flavio: You need it if you want to implement a custom Derive. However, it turns out we don't need it, more below

flavio:
@ereslibre I've updated the code to use derive the serialize and deserialize functions. I think everything looks fine right now, because our enums are unit-like (they are not tuples). I think the serialization output would change once we change that.

Because of that, I prefer to leave the previous commit into the history, plus leave the unit tests I added.

src/policy_evaluator.rs
flavio: Done

ereslibre (approved):
Thank you 🙏

PR #28, opened 2021-08-12
Refactor enum calls

ereslibre:
@flavio the first commit contains the trait approach that I worked with. I later realize that I didn't need to consume the trait itself anywhere and I cleaned it up in the second commit. If can squash if desired.


flavio (changes requested):
This looks much better. Please leave around the 1st commit, I'm fine with it.

Can you just do one final change and move the `BurregoRuntime` and the `WapcRuntime` code into dedicated files (one per runtime)?

This file is pretty big...

PR #29, opened 2021-08-13
Update Metadata validation rules

flavio:
The Metadata validation rules must take into account how the Opa/Gatekeeper policies behave.

The protocol version is allowed to be "unknown" for them.

### DIFF
--- a/src/policy_metadata.rs
+++ b/src/policy_metadata.rs
@@ -120,8 +120,9 @@ fn validate_resources(data: &[String]) -> Result<(), ValidationError> {
 
 #[derive(Deserialize, Serialize, Debug, Clone, Validate)]
 #[serde(rename_all = "camelCase")]
+#[validate(schema(function = "validate_metadata", skip_on_field_errors = false))]
 pub struct Metadata {
-    #[validate(required, custom = "validate_protocol_version")]
+    #[validate(required)]
     pub protocol_version: Option<ProtocolVersion>,
     #[validate]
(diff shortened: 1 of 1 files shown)


src/policy_metadata.rs
```diff
@@ -164,13 +165,15 @@ impl Metadata {
     }
 }
 
-fn validate_protocol_version(protocol: &ProtocolVersion) -> Result<(), ValidationError> {
-    match protocol {
-        ProtocolVersion::Unknown => Err(ValidationError::new(
+fn validate_metadata(metadata: &Metadata) -> Result<(), ValidationError> {
+    if metadata.execution_mode == PolicyExecutionMode::KubewardenWapc
+        && metadata.protocol_version == Some(ProtocolVersion::Unknown)
```
ereslibre, line 170: Should we consider the case when `protocolVersion` is `None` to error out as well?

ereslibre (approved):
LGTM. A question regarding the case when the metadata lacks some information. Not sure if that case can even happen right now.

src/policy_metadata.rs
flavio: This is already done, I've added a `#[validate(required)]` macro for it (like 125).

This crate we're using works in this way:
  1. Ensure all the field-specific validations are passing
  2. Ensure the "global" validation against the full object passes (that's the "schema function")

ereslibre: That's super nice, thanks for the information!

Issue #23, opened 2021-08-10
Burrego: allow end user to provide code for certain OPA functions


flavio:
Certain "host side" OPA functions should not have a hard coded implementation, but rather allow the end user to specify what has to be done when the function is invoked.

These functions should have implementations provided by the end user:

* `opa_abort`
* `opa_println`

ereslibre:
Implemented in https://github.com/kubewarden/policy-evaluator/pull/30

PR #30, opened 2021-08-16
User provided host functions

ereslibre:
Fixes https://github.com/kubewarden/policy-evaluator/issues/23

### DIFF
--- a/crates/burrego/src/opa/default_host_callbacks.rs
+++ b/crates/burrego/src/opa/default_host_callbacks.rs
@@ -1,12 +1,12 @@
 pub(crate) struct DefaultHostCallbacks;
 
 impl DefaultHostCallbacks {
-    pub(crate) fn opa_abort(msg: serde_json::Value) {
+    pub(crate) fn opa_abort(msg: String) {
         println!("OPA abort with message: {:?}", msg);
         std::process::exit(1);
     }
 
-    pub(crate) fn opa_println(msg: serde_json::Value) {
+    pub(crate) fn opa_println(msg: String) {
         println!("Message coming from the policy: {:?}", msg);
     }
 }
--- a/crates/burrego/src/opa/default_host_callbacks.rs
+++ b/crates/burrego/src/opa/default_host_callbacks.rs
@@ -1,0 +1,12 @@
+pub(crate) struct DefaultHostCallbacks;
+
+impl DefaultHostCallbacks {
+    pub(crate) fn opa_abort(msg: serde_json::Value) {
+        println!("OPA abort with message: {:?}", msg);
+        std::process::exit(1);
+    }
+
+    pub(crate) fn opa_println(msg: serde_json::Value) {
+        println!("Message coming from the policy: {:?}", msg);
+    }
+}
--- a/crates/burrego/src/opa/wasm.rs
+++ b/crates/burrego/src/opa/wasm.rs
@@ -127,7 +127,7 @@ impl Evaluator {
                         let msg = stack_helper
                             .pull_json(caller.as_context_mut(), &memory, addr)
                             .unwrap();
-                        (host_callbacks.$name)(msg);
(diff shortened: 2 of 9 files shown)


crates/burrego/src/opa/default_host_callbacks.rs
```diff
@@ -0,0 +1,12 @@
+pub(crate) struct DefaultHostCallbacks;
+
+impl DefaultHostCallbacks {
+    pub(crate) fn opa_abort(msg: serde_json::Value) {
+        println!("OPA abort with message: {:?}", msg);
+        std::process::exit(1);
+    }
+
+    pub(crate) fn opa_println(msg: serde_json::Value) {
+        println!("Message coming from the policy: {:?}", msg);
```
ereslibre, line 10: Do you think I should take care of doing `tracing` crate calls here?

flavio: I think the default implementation should just use println for that stuff, not tracing. I would leave the code as it is right now.

Note well: the default implementation should be something designed for the burrego cli program.

crates/burrego/src/opa/wasm.rs
```diff
@@ -110,36 +118,25 @@ impl Evaluator {
...
+                        let stack_helper = caller.data().unwrap();
+                        let msg = stack_helper
+                            .pull_json(caller.as_context_mut(), &memory, addr)
+                            .unwrap();
+                        (host_callbacks.$name)(msg);
+                    },
+                );
+                linker.define("env", stringify!($name), $name)?;
+            };
+        }
```
flavio, line 135: Wouldn't it be better to move that outside of the `new` function?

src/policy_evaluator.rs
```diff
@@ -181,7 +124,11 @@ impl PolicyEvaluator {
                     Policy::new,
                     policy_execution_mode,
                 )?;
-                let evaluator = burrego::opa::wasm::Evaluator::new(id, &policy_contents)?;
+                let evaluator = burrego::opa::wasm::Evaluator::new(
+                    id,
+                    &policy_contents,
+                    &opa_callbacks::DEFAULT_HOST_CALLBACKS,
```
flavio, line 130: Here we should not use the default host callbacks of burrego, but rather provide ours (which would using the `tracing` crate among other things).

src/policy_evaluator.rs
```diff
@@ -268,6 +215,8 @@ impl PolicyEvaluator {
 mod tests {
     use super::*;
 
+    use std::collections::HashMap;
```
flavio, line 218: Is that really needed, I don't see any change below compared to the original tests? :thinking:

flavio (changes requested):
Great contribution. I left some comments.

crates/burrego/src/opa/wasm.rs
ereslibre: It's closer to the usage of the macro (only the two next lines actually call it), and the macro captures variables from the lexical context, such as `store` and `memory`. They could be arguments to the macro but the locality of the macro still applies. I can move it if you think is more clear though.

src/policy_evaluator.rs
ereslibre: The `HashMap` itself of the Wapc mapping got moved to another mod. However, if I keep the `use` I get warnings of an unused import, because in this file, the `HashMap` is only used in `test` build mode. This is why the usage has to go inside the conditional compilation, so it builds and runs the unit tests, and it also does not print any warning about an unused import.

crates/burrego/src/opa/wasm.rs
flavio: Good point, no worries then

src/policy_evaluator.rs
flavio: Thanks!

flavio (approved):
LGTM :clap:

Issue #32, opened 2021-08-17
Implement `opa_println` for the burrego backend


ereslibre:
The burrego backend has two pluggable Wasm imports: this card is about implementing the `opa_println` function documented in https://www.openpolicyagent.org/docs/latest/wasm/#imports.

**Acceptance criteria**

- The result of this import call should be a tracing crate call that logs with `info` loglevel the provided message.

Issue #31, opened 2021-08-17
Implement `opa_abort` for the burrego backend


ereslibre:
The burrego backend has two pluggable Wasm imports: this card is about implementing the `opa_abort` function documented in https://www.openpolicyagent.org/docs/latest/wasm/#imports.

**Acceptance criteria**

- The result of this import call should be an HTTP 500 error with a message containing the error reported by the policy.

PR #34, opened 2021-08-18
Implement callbacks for `opa_abort` and `opa_println` Wasm imports

ereslibre:
Fixes: https://github.com/kubewarden/policy-evaluator/issues/31
Fixes: https://github.com/kubewarden/policy-evaluator/issues/32

**Note**: I could not find a way to produce Wasm policies that take advantage of this calls. I'm not sure if they are documented because the OPA team wants to implement them in the future, but I could not find a way to produce such a policy so I can test and add tests that validate the implementation.

Implement host callbacks for `opa_abort` and `opa_println` Wasm
imports that Rego policies might invoke when built for the Wasm
target.

- In the `burrego` CLI:

  - `abort` directly calls to `std::process::exit`, aborting the
  program execution.

  - `println` is mapped to what the `println!` macro expands to.

- In the `burrego` runtime backend for the `policy-evaluator`.

  - `abort` sets that the policy has requested abortion on the store,
  and when returning the result, an error is returned. This is doable
  because of the nature of Rego not being Turing complete. Otherwise,
  aborting the execution of the Wasm module on the wasmtime runtime
  environment would be required. In any case, aborting the execution
  would be nice to have and is left as a TODO comment.

  - `println` relies on the `tracing` crate.

### DIFF
--- a/crates/burrego/src/opa/default_host_callbacks.rs
+++ b/crates/burrego/src/opa/default_host_callbacks.rs
@@ -1,5 +1,16 @@
 pub(crate) struct DefaultHostCallbacks;
 
+use crate::opa::host_callbacks::HostCallbacks;
+
+impl Default for HostCallbacks {
+    fn default() -> HostCallbacks {
+        HostCallbacks {
+            opa_abort: Box::new(DefaultHostCallbacks::opa_abort),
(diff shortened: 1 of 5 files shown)


ereslibre:
Superseded by https://github.com/kubewarden/policy-evaluator/pull/35

PR #35, opened 2021-08-19
Opa callbacks

flavio:
This PR is built on top of https://github.com/kubewarden/policy-evaluator/pull/34 and supersede it.

Differences introduced by this PR:

* `opa_abort`
  * Fix how the error message of `opa_abort` is retrieved from the Wasm memory
  * Do not cause the running process to exit immediately with an error. An `Err` is automatically raised by Wasmtime. The same Wasmtime stack (engine, instance, memory,...) can be reused to process more requests

### DIFF
--- a/crates/burrego/src/opa/mod.rs
+++ b/crates/burrego/src/opa/mod.rs
@@ -51,7 +51,7 @@ impl StackHelper {
     /// * The data read
     pub fn read_string(
         &self,
-        store: impl AsContextMut,
+        store: impl AsContext,
         memory: &Memory,
         addr: i32,
     ) -> Result<Vec<u8>> {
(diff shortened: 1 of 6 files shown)


flavio:
@ereslibre you can start reviewing it, but I have now to look into the println handling

crates/burrego/src/opa/mod.rs
```diff
@@ -42,22 +42,22 @@ impl StackHelper {
...
     /// * `memory` - the Wasm linear memory used by the Wasm Instance
     /// * `addr` - address inside of the Wasm linear memory where the value is stored
     /// # Returns
-    /// * The JSON data read
-    pub fn pull_json(
+    /// * The data read
+    pub fn read_string(
         &self,
-        mut store: impl AsContextMut,
+        store: impl AsContextMut,
```
ereslibre, line 54: Either

```suggestion
        mut store: impl AsContextMut,
```

or

```suggestion
        store: impl AsContext,
```

ereslibre (approved):

Issue #38, opened 2021-08-20
burrego: implement missing Regex builtins


flavio:
These are the builtins that burrego has to provide that belong to the [Regex](https://www.openpolicyagent.org/docs/latest/policy-reference/#regex) area:

* [ ] `regex.split`
* [ ] `regex.globs_match`
* [ ] `regex.template_match`
* [ ] `regex.find_n`

PR #57, opened 2021-09-01
Implement regex builtins

ereslibre:
Related: https://github.com/kubewarden/policy-evaluator/issues/38

Implements the following builtins:

- `regex.split`
- `regex.template_match`
- `regex.find_n`

From this group, `regex.globs_match` is left unimplemented due to the
ratio effort/benefit and lack of real world consumers.


flavio (approved):
Wow, that wasn't trivial. :clap:

Issue #39, opened 2021-08-20
burrego: implement missing Glob builtins


flavio:
Epic: https://github.com/kubewarden/policy-evaluator/issues/56

These are the builtins that burrego has to provide that belong to the [Glob](https://www.openpolicyagent.org/docs/latest/policy-reference/#glob) area:

* [x] `glob.quote_meta`

ereslibre:
Implemented.

PR #58, opened 2021-09-01
Implement glob builtins

ereslibre:
Fixes: https://github.com/kubewarden/policy-evaluator/issues/39

### DIFF
--- a/crates/burrego/src/opa/builtins/glob.rs
+++ b/crates/burrego/src/opa/builtins/glob.rs
@@ -1,0 +1,46 @@
+use anyhow::{anyhow, Result};
+
+pub fn quote_meta(args: &[serde_json::Value]) -> Result<serde_json::Value> {
+    if args.len() != 1 {
+        return Err(anyhow!("glob.quote_meta: wrong number of arguments"));
+    }
+
+    let input = args[0]
+        .as_str()
+        .ok_or_else(|| anyhow!("glob.quote_meta: 1st parameter is not a string"))?;
+
+    serde_json::to_value(escape(input))
+        .map_err(|e| anyhow!("glob.quote_meta: cannot convert value into JSON: {:?}", e))
+}
+
+fn escape(s: &str) -> String {
+    let mut escaped = String::new();
+    for c in s.chars() {
+        match c {
+            '*' | '?' | '\\' | '[' | ']' | '{' | '}' => {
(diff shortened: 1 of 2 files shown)


crates/burrego/src/opa/builtins/glob.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    serde_json::to_value(escape(input))
+        .map_err(|e| anyhow!("glob.quote_meta: cannot convert value into JSON: {:?}", e))
+}
+
+fn escape(s: &str) -> String {
+    let mut escaped = String::new();
+    for c in s.chars() {
+        match c {
+            '?' | '*' | '[' | ']' => {
```
flavio, line 20: Looking at the `glob.match` documentation I think we might have to handle also: `:`, `{` and `}`.

I haven't looked at the source code of opa though...

flavio (commented):
I left one question.

As a final note, this doesn't implement `glob.match`, hence we cannot close https://github.com/kubewarden/policy-evaluator/issues/39

crates/burrego/src/opa/builtins/glob.rs
ereslibre: The go implementation uses https://github.com/gobwas/glob. I'm going to expand the quote so we match all special characters.

ereslibre: Their implementation check the glob special chars and escapes by prepending a `\`: https://github.com/open-policy-agent/opa/blob/581ba9c456be551b584759ae91f99ae6ec6b347b/vendor/github.com/gobwas/glob/glob.go#L68-L77.

The list of special characters is determined by: https://github.com/open-policy-agent/opa/blob/581ba9c456be551b584759ae91f99ae6ec6b347b/vendor/github.com/gobwas/glob/syntax/lexer/lexer.go#L23-L35.

I'm just copy-catting this behavior here, updated the PR.

ereslibre:
> As a final note, this doesn't implement glob.match, hence we cannot close #39

We can close it, this is about `glob` builtins. https://github.com/kubewarden/policy-evaluator/pull/57 was targeting `regex` builtins, and in that case, that issue had `References:` instead of `Fix:` to signal exactly that.

However, `glob.quote_meta` is the only builtin for the `glob` group, and we are good by closing #39 after this merges.

PR #64, opened 2021-09-22
Abort if builtin missing

ereslibre:
Fixes: https://github.com/kubewarden/kwctl/issues/63
Related: https://github.com/kubewarden/policy-server/issues/86

This issue is not marked as fixing https://github.com/kubewarden/policy-server/issues/86 because we could go the extra mile in the case of the policy-server and in case of several policies with missing builtins report all of them at once.

I can open a PR targeting the policy-server specifically after this one merges.

Also: added an extra commit to only register wapc policies in the wapc mapping.


ereslibre:
This is how it's reported with `kwctl`:

```
$ kwctl run -e gatekeeper --settings-json '{"labels":[{"key":"owner-team"}]}' --request-path ~/gatekeeper/missing-label-deployment.json ~/gatekeeper/policy.wasm
Error: error creating policy evaluator for policy file:///home/ereslibre/gatekeeper/policy.wasm: missing Rego builtins: rego.parse_module. Aborting execution.
```

flavio:
> This issue is not marked as fixing kubewarden/policy-server#86 because we could go the extra mile in the case of the policy-server and in case of several policies with missing builtins report all of them at once.
>
> I can open a PR targeting the policy-server specifically after this one merges.

Yes, I think it would be better to have a full list of all the policies that cannot be loaded

flavio:
> Also: added an extra commit to only register wapc policies in the wapc mapping.

Nice catch

flavio (approved):
LGTM

PR #67, opened 2021-11-16
Expose "get OCI manifest digest" callback

flavio:
Expose the get OCI manifest digest callback

Allow the Wasm guests to obtain the manifest's digest of
an OCI object (be it a container or an artifact).

Fetching information from a OCI registry requires the execution
of asynchronous code. Unfortunately the `host_callback` function
is a synchronous function.

Creating a dedicated tokio runtime (even by reusing the "current thread"
of the `host_callback` function) is not going to work. This would cause
tokio to cause a `panic` at runtime.

The solution of this problem consists of using a tokio channel that can
send the evaluation request from the synchronous world (the `host_callback`
function) to an `asynchronous` "worker" (something created with a tokio
`spawn` for example).

The response from the asynchronous world is then sent back to the
synchronous one by using a dedicated tokio `oneshot` channel.

## Changes

On top of exposing the new waPC host function, and adding all the tokio
channels. The following changes have been done:

  * The creation of a `PolicyEvaluator` is now done by using a dedicated
    builder: `PolicyEvaluatorBuilder`.
  * The code executing `PolicyEvaluator` (kwctl, policy-server) must
    prepare the worker that can handle the evaluation requests coming
    from the waPC guests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,11 @@ k8s-openapi = { version = "0.13.0", default-features = false, features = ["v1_22
 kube = { version = "0.64.0", default-features = false, features = ["client", "rustls-tls"] }
 kubewarden-policy-sdk = "0.2.2"
 lazy_static = "1.4.0"
+oci-distribution = { git = "https://github.com/krustlet/oci-distribution", rev = "96f19570e7f9dd56ed9cce8c9e8095e7c2d7753b" }
+policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.2.2" }
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
+tokio = { version = "^1", features = ["rt", "rt-multi-thread"] }
 tracing = "0.1"
 tracing-futures = "0.2"
 validator = { version = "0.13", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ members = [
 anyhow = "1.0"
 base64 = "0.13.0"
 burrego = { path = "crates/burrego" }
+cached = "0.26.2"
 hyper = { version = "0.14" }
 json-patch = "0.2.6"
 k8s-openapi = { version = "0.13.0", default-features = false, features = ["v1_22"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "policy-evaluator"
-version = "0.2.5"
+version = "0.2.6"
 authors = [
   "Flavio Castelli <fcastelli@suse.com>",
   "Rafael Fernández López <rfernandezlopez@suse.com>"
--- a/src/callback_handler/mod.rs
+++ b/src/callback_handler/mod.rs
@@ -104,6 +104,11 @@ impl CallbackHandler {
     pub async fn loop_eval(&mut self) {
         loop {
             tokio::select! {
+                // place the shutdown check before the message evaluation,
+                // as recommended by tokio's documentation about select!
+                _ = &mut self.shutdown_channel => {
+                    return;
+                },
                 maybe_req = self.rx.recv() => {
                     if let Some(req) = maybe_req {
                         match req.request {
@@ -129,9 +134,6 @@ impl CallbackHandler {
                         }
                     }
                 },
-                _ = &mut self.shutdown_channel => {
-                    return;
-                }
             }
         }
     }
--- a/src/callback_handler/mod.rs
+++ b/src/callback_handler/mod.rs
@@ -1,0 +1,133 @@
+use anyhow::{anyhow, Result};
+use policy_fetcher::{registry::config::DockerConfig, sources::Sources};
+use tokio::sync::{mpsc, oneshot};
+use tracing::warn;
+
+use crate::callback_requests::{CallbackRequest, CallbackRequestType, CallbackResponse};
+
+mod oci;
+
+const DEFAULT_CHANNEL_BUFF_SIZE: usize = 100;
+
+/// Helper struct that creates CallbackHandler objects
+pub struct CallbackHandlerBuilder {
+    oci_sources: Option<Sources>,
+    docker_config: Option<DockerConfig>,
+    channel_buffer_size: usize,
+    shutdown_channel: Option<oneshot::Receiver<()>>,
+}
+
+impl Default for CallbackHandlerBuilder {
+    fn default() -> Self {
+        CallbackHandlerBuilder {
+            oci_sources: None,
+            docker_config: None,
+            shutdown_channel: None,
+            channel_buffer_size: DEFAULT_CHANNEL_BUFF_SIZE,
+        }
+    }
+}
+
+impl CallbackHandlerBuilder {
+    #![allow(dead_code)]
+
+    /// Provide all the information needed to access OCI registries. Optional
+    pub fn registry_config(
+        mut self,
+        sources: Option<Sources>,
+        docker_config: Option<DockerConfig>,
+    ) -> Self {
+        self.oci_sources = sources;
+        self.docker_config = docker_config;
+        self
+    }
+
+    /// Set the size of the channel used by the sync world to communicate with
+    /// the CallbackHandler. Optional
+    pub fn channel_buffer_size(mut self, size: usize) -> Self {
+        self.channel_buffer_size = size;
+        self
+    }
+
+    /// Set the onetime channel used to stop the endless loop of
+    /// CallbackHandler. Mandatory
+    pub fn shutdown_channel(mut self, shutdown_channel: oneshot::Receiver<()>) -> Self {
+        self.shutdown_channel = Some(shutdown_channel);
+        self
+    }
+
+    /// Create a CallbackHandler object
+    pub fn build(self) -> Result<CallbackHandler> {
+        let (tx, rx) = mpsc::channel::<CallbackRequest>(self.channel_buffer_size);
+        if self.shutdown_channel.is_none() {
+            return Err(anyhow!("shutdown_channel_rx not provided"));
+        }
+
+        let oci_client = oci::Client::new(self.oci_sources, self.docker_config);
+        Ok(CallbackHandler {
+            oci_client,
+            tx,
+            rx,
+            shutdown_channel: self.shutdown_channel.unwrap(),
+        })
+    }
+}
+
+/// Struct that computes request coming from a Wasm guest.
+/// This should be used only to handle the requests that need some async
+/// code in order to be fulfilled.
+pub struct CallbackHandler {
+    oci_client: oci::Client,
+    rx: mpsc::Receiver<CallbackRequest>,
+    tx: mpsc::Sender<CallbackRequest>,
+    shutdown_channel: oneshot::Receiver<()>,
+}
+
+impl CallbackHandler {
+    /// Returns the sender side of the channel that can be used by the sync code
+    /// (like the `host_callback` function of PolicyEvaluator)
+    /// to request the computation of async code.
+    ///
+    /// Can be invoked as many times as wanted.
+    pub fn sender_channel(&self) -> mpsc::Sender<CallbackRequest> {
+        self.tx.clone()
+    }
+
+    /// Enter an endless loop that:
+    ///    1. Waits for requests to be evaluated
+    ///    2. Evaluate the request
+    ///    3. Send back the result of the evaluation
+    ///
+    /// The loop is interrupted only when a message is sent over the
+    /// `shutdown_channel`.
+    pub async fn loop_eval(&mut self) {
+        loop {
+            tokio::select! {
+                maybe_req = self.rx.recv() => {
+                    if let Some(req) = maybe_req {
+                        match req.request {
+                            CallbackRequestType::OciManifestDigest {
+                                image,
+                            } => {
+                                let response =
+                                    self.oci_client
+                                        .digest(&image)
+                                        .await
+                                        .map(|digest| CallbackResponse {
+                                            payload: digest.as_bytes().to_vec(),
+                                        });
+
+                                if let Err(e) = req.response_channel.send(response) {
+                                    warn!("callback handler: cannot send response back: {:?}", e);
+                                }
+                            }
+                        }
+                    }
+                },
+                _ = &mut self.shutdown_channel => {
+                    return;
+                }
+            }
+        }
+    }
+}
--- a/src/callback_handler/mod.rs
+++ b/src/callback_handler/mod.rs
@@ -1,7 +1,8 @@
 use anyhow::{anyhow, Result};
+use cached::proc_macro::cached;
 use policy_fetcher::{registry::config::DockerConfig, sources::Sources};
 use tokio::sync::{mpsc, oneshot};
-use tracing::warn;
+use tracing::{debug, warn};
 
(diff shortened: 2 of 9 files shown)


flavio:
This is now ready for review. We have to wait for the new version of `policy-fetcher` to be tagged, before we can merge this code

src/callback_requests.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+/// Describes the different kinds of request a waPC guest can make to
+/// our host.
+#[derive(Debug)]
+pub enum CallbackRequest {
+    /// Require the computation of the manifest digest of an OCI object (be
+    /// it an image or anything else that can be stored into an OCI registry)
+    OciManifestDigest {
+        /// String pointing to the object (e.g.: `resitry.testing.lan/busybox:1.0.0`)
```
ereslibre, line 18:
```suggestion
        /// String pointing to the object (e.g.: `registry.testing.lan/busybox:1.0.0`)
```

src/callback_requests.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Debug)]
+pub struct CallbackResponse {
+    /// The data to be given back to the waPC guest
+    pub payload: Vec<u8>,
+}
+
+/// Describes the different kinds of request a waPC guest can make to
+/// our host.
+#[derive(Debug)]
+pub enum CallbackRequest {
```
ereslibre, line 14: What do you think about `CallbackRequest` being parameterized with the response type and having a two-level request, given every enum variant will have the response. Pseudorust:

```rust
pub enum CallbackRequestType {
  OciManifestDigest { image: String },
}

pub struct CallbackRequest<R> {
  request: CallbackRequestType,
  response_channel: oneshot::Sender<Result<R>>,
}
```

src/policy_evaluator.rs
```diff
@@ -110,27 +111,22 @@ impl fmt::Debug for PolicyEvaluator {
...
-        policy_file: &Path,
-        policy_execution_mode: PolicyExecutionMode,
-        settings: Option<serde_json::Map<String, serde_json::Value>>,
-    ) -> Result<PolicyEvaluator> {
-        PolicyEvaluator::from_contents(id, fs::read(policy_file)?, policy_execution_mode, settings)
-    }
-
-    pub fn from_contents(
+    /// This method should not be used directly. Please use
+    /// `PolicyEvaluatorBuilder` instead.
```
ereslibre, line 115: We can remove this comment given it's not part of public API

src/policy_evaluator_builder.rs
```diff
@@ -0,0 +1,106 @@
+use anyhow::{anyhow, Result};
+use std::{fs, path::Path};
+use tokio::sync::mpsc::Sender;
+
+use crate::callback_requests::CallbackRequest;
+use crate::policy_evaluator::{PolicyEvaluator, PolicyExecutionMode};
+
+/// Helper Struct that creates a `PolicyEvaluator` object
+#[derive(Default)]
+pub struct PolicyEvaluatorBuilder {
```
ereslibre, line 10: I must confess I love this pattern in Rust.

src/runtimes/wapc.rs
```diff
@@ -44,6 +46,90 @@ pub(crate) fn host_callback(
...
                     Ok(Vec::new())
                 }
+                _ => {
+                    error!(namespace, operation, "unknown operation");
+                    Err(format!("unknown operation: {}", operation).into())
+                }
+            },
+            "oci" => match operation {
+                "manifest_digest" => {
+                    let policy_mapping = WAPC_POLICY_MAPPING.read().unwrap();
```
ereslibre, line 56: What is the expectation with burrego based policies? Do we plan to support something like this on those?

ereslibre (commented):
Huge step in a great direction. Love it.

src/policy_evaluator.rs
flavio: I left if on purpose for us, users of the internal `pub(crate)` method

src/runtimes/wapc.rs
flavio: I don't know, maybe in the future... but right now there's no way for them to reach out to the host. The only way would be to have us define some new Rego built-in and then plug that into the `CallbackHandler`

src/callback_requests.rs
flavio: This is a really good idea. I'll change that

flavio: Done

flavio:
I think this could be the last round of reviews

viccuad (approved):
LGTM!

Cargo.toml
```diff
@@ -24,8 +24,11 @@ k8s-openapi = { version = "0.13.0", default-features = false, features = ["v1_22
 kube = "0.64.0"
 kubewarden-policy-sdk = "0.2.2"
 lazy_static = "1.4.0"
+oci-distribution = { git = "https://github.com/kubewarden/krustlet", branch = "oci-distribution-fix-push-to-ghcr"}
+policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.2.0" }
```
ereslibre, line 28: I am only a bit concerned about policy-evaluator depending on policy-fetcher. This is not a blocker, but I would prefer that they don't depend on each other, maybe by splitting common behaviors in other crate that both depend on.

ereslibre (approved):
LGTM, just a comment regarding dependencies, but I think we can tackle at a later time.

Cargo.toml
flavio: The main issue we have is: we need to interact with a container registry, to do that we need a docker configuration file and a file defining the sources (what is insecure, what is using a self-signed certificate,...).

These data types (`DockerConfig`, `Sources`) are defined inside of policy-fetcher. This crate has also all the conversion functions that can turn these data into something that `oci-distribution` can understand.

I have to admit my head :boom: every time I have to do all these conversions... And I had to do basically the same thing with the sigstore-rs crate, in order to not expose the mutable (as in they could change over the time) `oci-distribution` types into my public API. It's just a gigantic mess...

Sorry about the rant, I really don't have a better idea about how to tackle that

ereslibre: It's fine, in this case I prefer to have it working, and see if we can refactor later.

flavio:
The last commit I pushed introduces the usage of a cache to temporarily store the results of the "get OCI manifest digest" invocations. 

I did that to address a significant performance drop inside of the PolicyServer.

# The testing scenario

I've ran our [load test](https://github.com/kubewarden/load-testing) against a local instance of PolicyServer. The PolicyServer was running uncontainerized and without TLS termination.

The PolicyServer was started with:
  * 2 worker threads to evaluate Wasm modules
  * 4 policies loaded, 1 of them used the "get OCI manifest digest" callback

The policy fetching an OCI manifest would reach a local docker registry running inside of my network.
The registry doesn't have authentication, but is TLS terminated using a certificate signed by trusted CA. 

## No policies using the OCI callback

To set a baseline, I configured the load tests to **not** reach the endpoint of the only policy using the OCI callback.

In this scenario the PolicyServer was able to process ~1500 requests per second.

## Hit the policy using the OCI callback - no caching

I updated the load test to ensure all the loaded policies, including the one using the OCI callback, could be hit.
It's important to stress that locust balances the request across all the different endpoints.

In this scenario the PolicyServer was able to process 15-20 requests per second (:exclamation: :exclamation: )

## Debugging the issue

@ereslibre and @viccuad raised some questions about the way the Callback Handler loop is managed inside of PolicyServer (see [this comment](https://github.com/kubewarden/policy-server/pull/126#discussion_r755281261)).

The CallbackHandler loop is not doing any blocking operation. It uses `await` and performs a`tokcio::select!` which is not causing the tokio runtime to block.

To confirm the performance issue is not caused by the CallbackHandler loop I did the following tests:
  * Performance are good when the CallbackHandler loop is running, but no request is sent to it. This is exactly the baseline scenario pictured before. The PolicyServer performance with the CallbackHandler loop running and without it (the code on the current `main` branch) are the same.
  * I replaced the code calculating the OCI digest to return a hard-coded string instead of reaching out to the actual registry. The load tests (hitting all the policies ofc) showed perfectly fine numbers

Thinking more about the issue, this is what actually happens:

* A POST request is received by the http server code of PolicyServer. This code runs inside of a tokio Runtime.
* The JSON request to be evaluated is sent from the API handler to the worker thread pool:
  * The communication between the async world (API handler) and the sync world (sync) happens over a tokio channel
  * The async world `await` on a tokio channel the response the sync world has to provide. This `await` operation is not blocking the tokio Runtime. This task will placed on pause and resumed as soon as a channel message is sent back to the async world

Thanks to async, the API can receive a really high number of request without getting stuck. The real limiting factor is inside of the sync world.

The worker pool is running inside of N+1 threads:
  * 1 thread is reserved to the code that receives all the requests to be evaluated. This is kinda like a phone operator from the old days
  * N threads: one per Wasm worker the user declared via the cli flag of PolicyServer

The "phone operator" from above works in this way:
  * It is always aware of which workers are busy processing a request and which ones are available
  * Once a message is received it finds the first available Wasm worker and dispatches to it the request to be evaluated

The "phone operator" is working in a non-blocking way. It never waits for the completion of a Wasm evaluation. It always reacts to events which can be of two types: an evaluation request is received, an evaluation result is ready.

Once the evaluation result is ready, the "phone operator" finds the channel associated with the API tokio task that generated the request and sends back the response of the tokio channel.

The API task is then resumed by the tokio runtime, the evaluation response is taken from the tokio channel and a HTTP response is sent back to the HTTP client that was waiting for it.

Having said that, the operation of "get the OCI manifest" involves two time expensive tasks:
  * A series of HTTP requests to fetch the manifest of an OCI image
  * Calculate the checksum of the manifest using the sha256 algorithm

These operation take some seconds, during which the Wasm worker is blocked waiting for an answer. This is a blocking operation, there's no other way to handle that because of how Wasm evaluation is done.
That means our "phone operator" can run temporarily run out of available Wasm workers if all of them are busy waiting for an OCI manifest to be calculated.

In our case, we had only 2 workers and 1 out of 4 policies causing these blocking "get me the OCI manifest" requests.

## The solution

There's no easy solution to this problem. Wasm evaluation cannot be done in an asynchronous way. The only workaround is to bite the bullet, but do that as seldom as possible.

The answer is to cache the result of the "get OCI manifest" operations, so that we pay this penalty a limited number of times.

This is what https://github.com/kubewarden/policy-evaluator/pull/67/commits/b6bcc6f67795ce255ec6eb36cae0c9cd83677896 does.

The cache we're using is time bound. Cached entries automatically expire after 60 seconds, and only successful results are cached. The latter point is a bit of a double sword: from one hand we don't want to a temporary hiccup with the network to cache an invalid results for 60 seconds, but from the other having a registry that is always down can cause a performance hit like the one we want to avoid.

## Testing the solution

I ran the load tests against a PolicyServer using this cache. I deliberately ran the code using a cache expiration of 10 seconds, instead of 60. I did that on purpose to simulate the impact of multiple cache misses.

The results I got have been super positive:

10 seconds cache

As you can see, we went back to almost normal numbers.

The numbers are even better with a 60 cache expiration:

60 seconds cache

## Overall

I think this is the way to go. It also makes sense IMO to cache this kind of value. I think 60 seconds is a reasonable amount of time.

viccuad:
:heart:  Awesome work! and I learned a fair share. So IIUC, problem is that the operations of HTTP requests & calculating checksum  take too long, and never yield from the await. With a finite worker pool in policy-server (all sharing threads for async/await), all workers are busy and no work can proceed timely.

I think a cache is perfect and you provided an awesome justification! 

I wonder on consuming this in the policy-server side. Looking at https://docs.rs/tokio/1.14.0/tokio/index.html#cpu-bound-tasks-and-blocking-code, I wonder if we should use `tokio::task::spawn_blocking()` for `loop eval()` so they run in their own thread in their own thread pool. I must confess that I haven't tried it, but I feel that it would solve the worker blocking issue, as they liberate their thread to run a different await.
In the case that the digest is not in cache (first setup, or bunch of resources being created at once), could all the workers spend a lot of time and essentially block? (I suppose if one uses `spawn_blocking()` the cache needs to be thread-safe, and the used [cached](https://docs.rs/cached/latest/cached/) already is. And you have set `#[cached(sync_writes = true)]`).

Also, I wonder if the shutdown_channel in `loop_eval()` can be called once we have the digest, for those other async calls that are asking for the same digest. Maybe this to much busy work for little payoff.

I also thought about non-blocking versions of the http client such as reqwest (we talked about that at some point?). But even with having concurrent http requests, the workers will spend too much time and block. And this doesn't solve the checksum calculation. So we need the cache.

flavio:
> So IIUC, problem is that the operations of HTTP requests & calculating checksum take too long, and never yield from the await. With a finite worker pool in policy-server (all sharing threads for async/await), all workers are busy and no work can proceed timely.

Not really. The HTTP/checksum operation are returning the control back to other tokio tasks. The problem is with the single wasm workers. they are not async (because wasmtime cannot work in this way), hence the workers are stuck in idle mode, waiting for the checkums to be returned to them.

> I wonder on consuming this in the policy-server side. Looking at https://docs.rs/tokio/1.14.0/tokio/index.html#cpu-bound-tasks-and-blocking-code, I wonder if we should use tokio::task::spawn_blocking() for loop eval() so they run in their own thread in their own thread pool. I must confess that I haven't tried it, but I feel that it would solve the worker blocking issue, as they liberate their thread to run a different await.

Based on what I wrote before, this would not help. What we would have to do is to allow the worker pool to grow dynamically based on the pressure we have on the queue. This is a future optimization we can do.

> In the case that the digest is not in cache (first setup, or bunch of resources being created at once), could all the workers spend a lot of time and essentially block? (I suppose if one uses spawn_blocking() the cache needs to be thread-safe, and the used cached already is. And you have set #[cached(sync_writes = true)]).

If two workers are asked to compute the same manifest digest and the cache is empty:
  * 1 worker will perform all the network/checksum requests
  * the other worker will wait for the value to be added to the cache, and then will resume operation

I think this approach is better than having two network requests done in parallel, 2 checksum computations done in parallel,...

> Also, I wonder if the shutdown_channel in loop_eval() can be called once we have the digest, for those other async calls that are asking for the same digest. Maybe this to much busy work for little payoff.

We have to let the loop eval run for as long as we are listening to the web API. The loop eval is not going to consume computation time. Quoting tokio's official docs:

```
> Tasks in Tokio are very lightweight. Under the hood, they require only a single allocation and 64 bytes of memory. Applications should feel free to spawn thousands, if not millions of tasks.
```

We need the loop to keep running because this is where all the "please compute this digest" messages that are sent by the Wasm workers are received and elaborated.

> I also thought about non-blocking versions of the http client such as reqwest (we talked about that at some point?). But even with having concurrent http requests, the workers will spend too much time and block. And this doesn't solve the checksum calculation. So we need the cache.

Exactly, you're spot on.

viccuad (approved):

src/callback_handler/mod.rs
```diff
@@ -131,3 +136,27 @@ impl CallbackHandler {
...
+    sync_writes = true,
+    key = "String",
+    convert = r#"{ format!("{}", img) }"#,
+    with_cached_flag = true
+)]
+async fn get_oci_digest_cached(
+    oci_client: &oci::Client,
+    img: &str,
+) -> Result<cached::Return<String>> {
+    oci_client.digest(img).await.map(cached::Return::new)
```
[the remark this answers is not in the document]
flavio: I would leave everything cached for a little bit... I'm too afraid of performance issues. This can always be tuned later on

flavio:
Merging. I just rebased a bit the history to make everything nicer

PR #68, opened 2021-11-17
Upgrade to the lastest version of tokio

flavio:
This is required to fix https://github.com/kubewarden/kwctl/issues/86

Currently in draft, the API of `kube` change and it breaks building


flavio:
I cannot review it because I'm the one who created the PR, but I approve it. Thanks for the help @ereslibre !

PR #71, opened 2021-12-31
Mutating field.

jvanz:
Adds the "mutating" field in the policy type to allow policy server checking if the policy evaluated is allowed to mutate the requests.

### DIFF
--- a/src/policy.rs
+++ b/src/policy.rs
@@ -14,6 +14,7 @@ use crate::callback_requests::CallbackRequest;
 #[derive(Clone)]
 pub struct Policy {
     pub id: String,
+    pub mutating: bool,
     policy_id: Option<u64>,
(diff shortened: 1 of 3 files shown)


flavio (changes requested):
Currently I'm not in favor of this change. I think we can avoid to do that and make the change only inside of the policy server (see my comments on https://github.com/kubewarden/policy-server/pull/139). But maybe I'm wrong and I didn't have enough :coffee: :)

ereslibre (changes requested):
I also think this behavior (and settings) belongs exclusively to the policy-server, I would not introduce this logic here.

flavio:
This is related with https://github.com/kubewarden/allow-privilege-escalation-psp-policy/issues/9

jvanz:
After the latest changes at https://github.com/kubewarden/policy-server/pull/139, this is not necessary anymore

PR #72, opened 2022-01-14
Bump policy SDK

ereslibre:


Cargo.toml
```diff
@@ -39,3 +39,6 @@ wasmtime-provider = "0.0.5"
 
 [dev-dependencies]
 assert-json-diff = "2.0.1"
+
+[patch.crates-io]
+kubewarden-policy-sdk = { git = "https://github.com/ereslibre/policy-sdk-rust", branch = "non-wasm" }
```
flavio, line 44: I think this is something we will have to get rid of

flavio (changes requested):
LGTM, aside from the minor detail I commented about

Cargo.toml
ereslibre: Yes, when https://github.com/kubewarden/policy-sdk-rust/pull/20 is merged and tagged, I will remove this patch, update the PR and tag a new release.

ereslibre:
Merging on green.

PR #115, opened 2022-04-13
Update kube crate to version v0.70.0

jvanz:
The kube crate version v0.70.0 added support for ECDSA algorithm. Which is necessary to fix a certificate issue when trying to run PolicyServer in K3D and Minikube cluster. This commit updates the crate version to fix the issue.

Related to https://github.com/kubewarden/policy-server/issues/136

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ burrego = { path = "crates/burrego" }
 cached = "0.30.0"
 hyper = { version = "0.14" }
 json-patch = "0.2.6"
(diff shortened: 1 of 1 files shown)


ereslibre (commented):
Thanks @jvanz! There is https://github.com/kubewarden/policy-evaluator/pull/114 that is bumping to a newer version. Can we bump to 0.71 if everything works?

jvanz:
> Thanks @jvanz! There is #114 that is bumping to a newer version. Can we bump to 0.71 if everything works?

Sure! I've just tested it with the v0.71.0 version. It works as expected. Closing this in favor of #114 .

PR #117, opened 2022-04-20
feat: Allow policy authors to write policies that perform image verification

raulcabello:
` v1/verify` operation added which can be called from the sdk to verify sigstore signatures for a given image. This new operation will receive a CallbackRequestType  which can be SigstorePubKeyVerify and SigstoreKeylessVerify. Verification results are cached for 60 seconds.

Fix https://github.com/kubewarden/policy-server/issues/99

### DIFF
--- a/src/callback_handler/mod.rs
+++ b/src/callback_handler/mod.rs
@@ -4,7 +4,9 @@ use policy_fetcher::{registry::config::DockerConfig, sources::Sources};
 use tokio::sync::{mpsc, oneshot};
 use tracing::{debug, warn};
 
-use crate::callback_requests::{CallbackRequest, CallbackRequestType, CallbackResponse};
+use crate::callback_requests::{CallbackRequest, CallbackResponse};
+
+use policy_fetcher::kubewarden_policy_sdk::host_capabilities::CallbackRequestType;
 
 mod oci;
 mod sigstore;
@@ -80,6 +82,7 @@ impl CallbackHandlerBuilder {
 /// code in order to be fulfilled.
 pub struct CallbackHandler {
     oci_client: oci::Client,
+    //sigstore_client: sigstore.Client,
     rx: mpsc::Receiver<CallbackRequest>,
     tx: mpsc::Sender<CallbackRequest>,
     shutdown_channel: oneshot::Receiver<()>,
@@ -134,8 +137,7 @@ impl CallbackHandler {
                             },
                             CallbackRequestType::SigstoreVerify{
                                 image: _,
-                                all_of: _,
-                                any_of: _,
+                                config: _,
                             } => todo!(),
                         }
                     }
--- a/src/callback_handler/mod.rs
+++ b/src/callback_handler/mod.rs
@@ -1,6 +1,6 @@
 use anyhow::{anyhow, Result};
 use cached::proc_macro::cached;
-use policy_fetcher::{registry::config::DockerConfig, sigstore, sources::Sources};
+use policy_fetcher::{registry::config::DockerConfig, sources::Sources};
 use std::collections::HashMap;
 use tokio::sync::{mpsc, oneshot};
 use tracing::{debug, warn};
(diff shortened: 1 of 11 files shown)


viccuad (approved):
LGTM :).

src/callback_handler/mod.rs
```diff
@@ -64,9 +72,18 @@ impl CallbackHandlerBuilder {
...
         }
 
-        let oci_client = oci::Client::new(self.oci_sources, self.docker_config);
+        let oci_client = oci::Client::new(self.oci_sources.clone(), self.docker_config.clone());
+        let repo = sigstore::tuf::SigstoreRepository::fetch(None)?;
+        let fulcio_and_rekor_data = FulcioAndRekorData::FromTufRepository { repo };
+        let sigstore_client = sigstore_verification::Client::new(
+            self.oci_sources.clone(),
+            self.docker_config.clone(),
+            &fulcio_and_rekor_data,
```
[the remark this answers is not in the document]
raulcabello: this is done in `policy-fetcher` https://github.com/kubewarden/policy-fetcher/pull/79/commits/72ee3e93ef965dde239d8e012badebd52f9cda2d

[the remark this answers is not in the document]
viccuad: I was confused, but after looking here, this `sigstore_client` is ancillary to policy-evaluator. And it calls a `policy-fetcher::Verifier:new()`, that is the one that creates the sigstore-rs client, and enables the cache. So all good from my side :).

viccuad (approved):
LGTM, approving :).

src/callback_handler/sigstore_verification.rs
```diff
@@ -57,9 +72,24 @@ impl Client {
         if keyless.is_empty() {
             return Err(anyhow!("Must provide keyless info"));
         }
+        // Build intering VerificationConfig:
```
flavio, line 75: Really silly comment, maybe it's a limitation of my English... but I don't get the meaning of this comment. :sweat_smile:

flavio (approved):
LGTM, I like how small the change turned out to be

src/callback_handler/sigstore_verification.rs
raulcabello: I believe @viccuad wrote this method (I just moved it from the fetcher to the evaluator) I think it means intermediate or auxiliary VerificationConfig, right Victor?

viccuad: yes, that's a typo, it should be "interim".

viccuad: Could be dropped for symmetry, it's not present in `verify_public_key()`.

flavio: do as you wish, I don't have any preference :)

raulcabello:
please review the last [commit](https://github.com/kubewarden/policy-evaluator/pull/117/commits/14d3308d0015de4add290f41afc26b5d81d53884) since we need that for https://github.com/kubewarden/kwctl/issues/205

src/callback_handler/mod.rs
```diff
@@ -71,14 +78,15 @@ impl CallbackHandlerBuilder {
         if self.shutdown_channel.is_none() {
             return Err(anyhow!("shutdown_channel_rx not provided"));
         }
+        if self.fulcio_and_rekor_data.is_none() {
+            return Err(anyhow!("fulcio_and_rekor_data not provided"));
+        }
```
flavio, line 83: This line, together with the one you added below don't match my **personal** style. Feel free to ignore this comment, I would however do something like that:

```rust
let fulcio_and_rekor_data = self.fulcio_and_rekor_data.ok_or_else(|| anyhow!("fulcio_and_rekor_data not provided"))?;

// later on
let sigstore_client = sigstore_verification::Client::new(
            self.oci_sources.clone(),
            self.oci_sources.clone(),
            self.docker_config.clone(),
            self.docker_config.clone(),
            &fulcio_and_rekor_data,
            fulcio_and_rekor_data,
        )?;
```

flavio (approved):
LGTM, I just left a minor comment about something that could also be ignored

src/callback_handler/mod.rs
raulcabello: I like that approach! it looks more idiomatic, and I don't like to see unwrap(). I changed also shutdown_channel, so it is done in the same way https://github.com/kubewarden/policy-evaluator/pull/117/commits/7327c5ab91a0338a0922a719a8e200698be754db

flavio (approved):
LGTM, thanks for the change

PR #118, opened 2022-04-29
Fix policy SDK dependency graph

flavio:
Do not consume the policy SDK via policy-fetcher, this is a leftover from one of the old iterations of the Sigstore integration.

Starting from v0.7.1 of policy-fetcher, the policy SDK is no longer re-exposed by policy-fetcher.

The policy SDK belongs to the policy-evaluator.

Depends on https://github.com/kubewarden/policy-fetcher/pull/83
I'll update this PR once the one against policy-fetcher gets merged

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,9 @@ cached = "0.30.0"
 hyper = { version = "0.14" }
 json-patch = "0.2.6"
 kube = { version = "0.71.0", default-features = false, features = ["client", "rustls-tls"] }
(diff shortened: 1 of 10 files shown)


raulcabello (approved):
LGTM, thanks!

flavio:
I've updated `Cargo.toml` to reference the release of policy-fetcher I just tagged

flavio:
@raulcabello : I've updated to policy-fetcher 0.7.2

PR #119, opened 2022-04-29
add wapc dns lookup callback

flavio:
This PR extents the waPC capabilities offered to our policy authors by introducing a new API that allows reverse DNS lookup.

This fixes https://github.com/kubewarden/policy-server/issues/236

While working on this PR, I also realized the "OCI get manifest invocations" were not encoding and decoding the waPC payload properly.

### DIFF
--- a/src/callback_handler/oci.rs
+++ b/src/callback_handler/oci.rs
@@ -1,4 +1,5 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
+use policy_fetcher::oci_distribution::Reference;
 use policy_fetcher::{registry::config::DockerConfig, registry::Registry, sources::Sources};
 
 /// Helper struct to interact with an OCI registry
@@ -15,9 +16,16 @@ impl Client {
 
     /// Fetch the manifest digest of the OCI resource referenced via `image`
     pub async fn digest(&self, image: &str) -> Result<String> {
-        let image_with_proto = format!("registry://{}", image);
-        self.registry
+        // this is needed to expand names as `busybox` into
+        // fully resolved references like `docker.io/library/busybox`
+        let image_ref: Reference = image.parse()?;
+
+        let image_with_proto = format!("registry://{}", image_ref.whole());
+        let image_digest = self
+            .registry
             .manifest_digest(&image_with_proto, self.sources.as_ref())
-            .await
+            .await?;
+        serde_json::to_string(&image_digest)
+            .map_err(|e| anyhow!("Cannot serialize response to json: {}", e))
     }
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "policy-evaluator"
-version = "0.3.1"
+version = "0.3.2"
 authors = [
   "Flavio Castelli <fcastelli@suse.com>",
   "Rafael Fernández López <rfernandezlopez@suse.com>"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,11 @@ anyhow = "1.0"
 base64 = "0.13.0"
 burrego = { path = "crates/burrego" }
 cached = "0.30.0"
+dns-lookup = "1.0.8"
 hyper = { version = "0.14" }
 json-patch = "0.2.6"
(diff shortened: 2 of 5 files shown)


viccuad (approved):
LGTM. We should add this new callback to the wapc protocol reference docs.

src/callback_handler/oci.rs
```diff
@@ -15,9 +16,16 @@ impl Client {
...
+        // fully resolved references like `docker.io/library/busybox`
+        let image_ref: Reference = image.parse()?;
+
+        let image_with_proto = format!("registry://{}", image_ref.whole());
+        let image_digest = self
+            .registry
             .manifest_digest(&image_with_proto, self.sources.as_ref())
-            .await
+            .await?;
+        serde_json::to_string(&image_digest)
```
raulcabello, line 28: why do we need this? `manifest_digest` returns a `Result<String>`. Can we just return `manifest_digest` ?

src/callback_handler/oci.rs
```diff
@@ -15,9 +16,16 @@ impl Client {
...
+            .map_err(|e| anyhow!("Cannot serialize response to json: {}", e))
```
raulcabello, line 29: why do we need this? `manifest_digest` returns a `Result<String>`. Can we just return `manifest_digest` ?

raulcabello (commented):
It looks good, I just left one comment

src/callback_handler/oci.rs
viccuad: I am doubting myself too here.
This change is consistent with the rest of callbacks, though. But in https://github.com/kubewarden/policy-evaluator/blob/main/src/callback_handler/mod.rs#L153, we use the digest as a vector of bytes, instead of doing a `serde_json::to_vec()`.

Maybe I'm misunderstanding things, but, for `CallbackRequestType::SigstoreKeylessVerify` and `CallbackRequestType::SigstorePubKeysVerify`, don't we transform the payload of the `CallbackResponse` to json twice? (so we create a json from a json)

raulcabello: ah actually I think we need it. `manifest_digest` returns a string, and `serde_json::to_string` will return a valid JSON for this String.
Then https://github.com/kubewarden/policy-evaluator/blob/main/src/callback_handler/mod.rs#L153 we just transform it to Vec<u8>, then we get the String back in the sdk.

raulcabello: > Maybe I'm misunderstanding things, but, for CallbackRequestType::SigstoreKeylessVerify and CallbackRequestType::SigstorePubKeysVerify, don't we transform the payload of the CallbackResponse to json twice? (so we create a json from a json). In the past I though that it was because we want to always return a json object, that wraps the response. Hence why we did it here for the digest. Am I understanding it correctly?

I think we just transform it to JSON once we are about to send it. [Here](https://github.com/kubewarden/policy-evaluator/blob/main/src/callback_handler/sigstore_verification.rs#L56) we just return a `VerificationResponse`, then we convert it to json [here](https://github.com/kubewarden/policy-evaluator/blob/main/src/callback_handler/mod.rs#L195)

raulcabello: Resolving if you agree @viccuad? since I now think this is fine, sorry about the noise

raulcabello (approved):

src/callback_handler/oci.rs
viccuad: Yes, I see it now. Thanks!

ereslibre (commented):
LGTM besides versioning.

Nit: I would remove the "reverse" part of the commit description, as I consider reverse lookup as `dig -x` as opposed to a regular `dig` (what this is implementing).

jvanz (approved):

flavio:
* rebased against master
* added versioning to waPC calls
* reworded the commit message as requested by rafa

viccuad (approved):

flavio:
One last review, this time to ensure the new response objects are being used. Once the new rust-sdk is released I'll update the Cargo.toml reference

flavio:
Changed `Cargo.toml`:
  * Point to the freshly released version of the Rust SDK
  * Bump policy-evaluator patch version: v0.3.2

I'm going to merge and tag a new release once all the tests are green. Thanks for the reviews

PR #143, opened 2022-06-20
use admission response

flavio:
Rename the `ValidationResponse` object to be `AdmissionResponse`. This uses the same terminology of Kubernetes. This is needed later on, inside of policy-server, when building proper AdmissionResponse `Response` objects.

While using the `AdmissionResponse`, make use of the fields that have been just exposed by the Rust policy SDK. This ensures we can create any kind of AdmissionResponse that Kubernetes supports.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "policy-evaluator"
-version = "0.3.7"
+version = "0.4.0"
 authors = [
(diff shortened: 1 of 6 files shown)


raulcabello (approved):
LGTM

flavio:
force pushed to use the freshly tagged SDK crate

PR #146, opened 2022-06-24
 feat: Support verifying GHA signatures

viccuad:
## Description

Add `verify_github_actions()`, which is called when we receive the new
`CallbackRequestType::SigstoreGithubActionsVerify{}`.

Depends on https://github.com/kubewarden/policy-sdk-rust/pull/46


raulcabello (approved):
LGTM, thanks! just don't forget to change `kubewarden-policy-sdk` in the `Cargo.toml`

crates/burrego/src/opa/builtins/regex.rs
```diff
@@ -117,7 +118,7 @@ impl Display for ExpressionList {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut result = String::new();
```
raulcabello, line 119: do we really need this intermediate String? Can we just write in the formatter inside the loop? something like:

```
fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        for expression in self.0.iter() {
            write!(f, "{}", expression)?;
        }
        Ok(())
    }
```
Then we wouldn't need this line ` use std::fmt::Write as _; // import without risk of name clashing`

flavio: :+1: with @raulcabello . If you want we can also tackle that with another PR, I'm fine merging that with the clippy warning about burrego

viccuad: agh, thanks, true! done in https://github.com/kubewarden/policy-evaluator/pull/146/commits/03c8a36ebbb30a7568957d1b592510af1b2e2297

crates/burrego/src/opa/builtins/regex.rs
```diff
@@ -117,9 +117,9 @@ impl Display for ExpressionList {
...
         for expression in self.0.iter() {
-            result.push_str(&format!("{}", expression));
+            write!(f, "{}", expression)?;
         }
-        write!(f, "{}", result)
+        Ok()
```
raulcabello, line 122:
```suggestion
        Ok(())
```
Does it compile just with Ok()? I think it should be Ok(()). We can't see it in the CI as it failed because policy-sdk 0.6.1 was not available at that time

viccuad: Indeed you are right, whoops. Amended.

raulcabello (approved):

flavio (approved):
ship it!

jvanz (approved):

PR #152, opened 2022-07-05
Consume SDK 0.6.2 for `v2/verify`

viccuad:
## Description

Consume deambiguation of KeylessPrefix request.
Support v2/verify callbacks.

### DIFF
--- a/src/runtimes/wapc.rs
+++ b/src/runtimes/wapc.rs
@@ -54,7 +54,7 @@ pub(crate) fn host_callback(
                 }
             },
             "oci" => match operation {
-                "v1/verify" => {
+                "v1/verify" | "v2/verify" => {
                     let req_type: CallbackRequestType =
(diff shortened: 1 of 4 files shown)


src/runtimes/wapc.rs
```diff
@@ -65,7 +65,7 @@ pub(crate) fn host_callback(
 
                     send_request_and_wait_for_response(policy_id, binding, operation, req, rx)
                 }
-                "v1/manifest_digest" => {
+                "v2/manifest_digest" => {
```
raulcabello, line 68: why we need to bump to v2 here? Isn't it enough just in verify?

viccuad: true! fixed in https://github.com/kubewarden/policy-evaluator/pull/152/commits/ae70eaf5a84e46fa9747a2a65f7619547307845c, rebased against sdk 0.6.2.

raulcabello (approved):

PR #160, opened 2022-07-18
Allow fulcio and rekor data to be None

flavio:
Do not fail when Fulcio and Rekor data are None. The Sigstore integration will still work, but it will provide some false negatives.

False netative means: a perfectly valid signature produced with keyless mode will be reported as non valid.

This is currently required to handle the case when the upstream TUF repository of Sigstore is broken.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "policy-evaluator"
-version = "0.4.4"
+version = "0.4.5"
 authors = [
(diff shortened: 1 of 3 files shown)


viccuad (approved):
LGTM, besides consuming the upcoming policy-fetcher version.

raulcabello (approved):

flavio:
> LGTM, besides consuming the upcoming policy-fetcher version.

Fixed, now that policy-fetcher has been tagged

PR #170, opened 2022-08-08
feat: use docker_credential crate instead of DockerConfig

### LINKED ISSUE #169: Remove DockerConfig and use docker_credential crate instead
Adapt the code and consume latest [changes](https://github.com/kubewarden/policy-fetcher/pull/111) from policy-fetcher where docker_credential crate is used.

raulcabello:
Fix #169

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ kube = { version = "0.74.0", default-features = false, features = ["client", "ru
 k8s-openapi = { version = "0.15.0", default-features = false }
 kubewarden-policy-sdk = "0.6.3"
 lazy_static = "1.4.0"
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.7.9" }
+policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.7.11" }
 serde_json = "1.0"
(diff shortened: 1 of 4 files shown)


flavio (changes requested):
Overall OK, but this is not introducing the changes required to keep the cli flag and the env variable that are used to set the docker configuration file to work.

raulcabello:
I was going to add the env var in [kwctl](https://github.com/kubewarden/kwctl/blob/main/src/main.rs#L415) and [policy-server](https://github.com/kubewarden/policy-server/blob/main/src/cli.rs#L282)

flavio:
> I was going to add the env var in [kwctl](https://github.com/kubewarden/kwctl/blob/main/src/main.rs#L415) and [policy-server](https://github.com/kubewarden/policy-server/blob/main/src/cli.rs#L282)

my fault... I misread `policy-evaluator` with `policy-server` :facepalm:

flavio (approved):
LGTM, make sure you consume the right version of policy-fetcher

flavio (approved):
nit, can you please squash the 2 commits together? I would prefer to not leave traces inside of the history of your policy-fetcher fork

jvanz (approved):
LGTM! :D

PR #186, opened 2022-09-23
Groundwork for Policy Server refactoring

flavio:
This PR paves the way to some refactoring inside of Policy Server which will allow us to put a limit on the evaluation time of policies.

These are the major themes of this PR:

- fix: improve performance of Metadata parsing
- refactoring/fix: bump burrego to version 0.2: lots of cleanup of the wasmtime-related code, plus reorganization of the codebase. The cleanup allows us to get rid of some global variables that were pretty ugly
- refactoring: small changes to PolicyEvaluatorBuilder. This is probably the most important part of the PR, because these changes allow some optimizations to take place inside of Policy Server


flavio:
Note: this depends on https://github.com/wapc/wapc-rs/pull/14

src/policy_evaluator_builder.rs
```diff
@@ -74,41 +100,223 @@ impl PolicyEvaluatorBuilder {
...
+
+                let evaluator = burrego::Evaluator::from_engine_and_module(
+                    engine,
+                    module,
+                    crate::runtimes::burrego::new_host_callbacks(),
+                )?;
+
+                let policy_runtime = Runtime::Burrego(Box::new(BurregoEvaluator {
+                    evaluator,
+                    entrypoint_id: 0, // currently hardcoded to this value
```
raulcabello, line 221: why hardcoded to 0? if we don't need it at this point should we just use the default value?

src/policy_evaluator_builder.rs
```diff
@@ -45,6 +63,14 @@ impl PolicyEvaluatorBuilder {
         self
     }
 
+    /// Use a pre-built [`wasmtime::Module`] instance.
+    /// **Warning:** you must provide also the [`wasmtime::Engine`] used
+    /// to allocate the `Module`, otherwise the code will panic at runtime
```
raulcabello, line 68: this will be checked in `validate_user_input`, and then it would return an error in `build`. I don't see how this can panic in the `policy-evaluator`. If it panics when it is called in the `policy-server` it might makes more sense to move the comment there

raulcabello (approved):
LGTM, thanks! I tested this with the new policy-server and the startup time is much faster!
I just left some minor comments

src/policy_evaluator_builder.rs
flavio: I prefer to leave it there, because this is something any kind of consumer of `PolicyEvaluatorBuilder` must know.

The check inside of `validate_user_input` cannot catch all kind of misuses. I'm referring to something like:

* User creates a `wasmtime::Engine` called `eng1`, which has the "fuel" feature enabled
* User creates a `wasmtime::Module` using `eng1`. The module will be compiled to have the fuel instructions added
* User uses `PolicyEvaluatorBuilder` to create a `PolicyEvaluator`. At creation time the user provides the module created previously and an engine via `wasmtime::Engine::new`

In this scenario, `PolicyEvaluatorBuilder::validate_user_input` will spot 0 issues. But at runtime, the code is going to panic because the engine being used has a different configuration from the one that compiled the module

src/policy_evaluator_builder.rs
flavio: We cannot implement the `Default` trait against `BurregoEvaluator`, there's no way to have a default `burrego::Evaluator`

We could move the hard-coded `0` into `BurregoEvaluator` here:
https://github.com/kubewarden/policy-evaluator/blob/d73d937366db297328d1cb9452b5eebe3849a2b4/src/runtimes/burrego.rs#L68-L71

However, it wouldn't be that different from the code you just commented on

Do you have other ideas?

jvanz (approved):

viccuad (approved):

src/policy_evaluator_builder.rs
viccuad: would hardcoding it there simplify much? AFAIU, the other runtimes have an `entrypoint_id` too?

flavio:
A new release of wapc/wasmtime-provider has been tagged :partying_face:

PR #213, opened 2022-11-24
feat: implement sigstore cert verification

flavio:
Allow sigstore based verification to be done using a certificate provided by the user.

This is required to implement: https://github.com/kubewarden/verify-image-signatures/issues/39

**Note:** I'm currently using code from git branches, this is going to be changed as soon as everything gets merged

### DIFF
--- a/src/callback_handler/sigstore_verification.rs
+++ b/src/callback_handler/sigstore_verification.rs
@@ -2,12 +2,21 @@ use anyhow::{anyhow, Result};
 use kubewarden_policy_sdk::host_capabilities::verification::{
     KeylessInfo, KeylessPrefixInfo, VerificationResponse,
 };
+use policy_fetcher::sigstore;
 use policy_fetcher::sources::Sources;
 use policy_fetcher::verify::config::{LatestVerificationConfig, Signature, Subject};
-use policy_fetcher::verify::{FulcioAndRekorData, Verifier};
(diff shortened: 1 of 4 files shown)


viccuad (approved):
LGTM

src/callback_handler/sigstore_verification.rs
```diff
@@ -16,8 +25,60 @@ impl Client {
         sources: Option<Sources>,
         fulcio_and_rekor_data: Option<&FulcioAndRekorData>,
     ) -> Result<Self> {
-        let verifier = Verifier::new(sources, fulcio_and_rekor_data)?;
-        Ok(Client { verifier })
+        let cosign_client = Arc::new(Mutex::new(Self::build_cosign_client(
```
viccuad, line 28: just a question. I see the usefulness of `build_cosign_client()`, but why do we need the cosign_client thread safe? I suppose I'm missing something obvious.

flavio: It's not that obvious, that's not a silly question! :)

It has to be thread safe because it's a mutable object, that is shared via pointers inside between the `Verifier` and the `Client`. Rust borrow checker rightfully complained about that. That was the only way to allow the sharing between the two objects

flavio:
Updated `Cargo.toml` to reference official releases instead of forks

PR #214, opened 2022-11-30
fix!: Make `verify_certificate()`, return `Result<BoolWithReason>`, & bump version

viccuad:
## Description

Relates to https://github.com/kubewarden/verify-image-signatures/pull/47#discussion_r1034836373

Make `verify_certificate()`, return `Result<policy-evaluator::BoolWithReason>`. Response either returns:
* `Ok(BoolWithReason::True)`.
* `Ok(BoolWithReason::False((reason))` with ``reason` such as "cert not trusted by cert
  chain", or "cert expired".
* `Err(e)`, with `e` such as "cert not in PEM format", or "not_after
  datetime not in RFC 3339 format".

Bump version to 0.4.17

## Test

Unit tests amended.

### DIFF
--- a/src/callback_handler/crypto.rs
+++ b/src/callback_handler/crypto.rs
@@ -11,10 +11,17 @@ struct CertificatePool {
     intermediates: Vec<picky::x509::Cert>,
 }
 
+#[derive(Debug)]
+/// Used as return of verify_certificate()
+pub enum BoolWithReason {
+    True,
+    False(String),
+}
+
 /// verify_certificate verifies the validity of the certificate, and if it is
 /// trusted with the provided certificate chain.
 /// If the provided certificate chain is empty, it is treated as trusted.
-pub fn verify_certificate(req: CertificateVerificationRequest) -> Result<bool> {
+pub fn verify_certificate(req: CertificateVerificationRequest) -> Result<BoolWithReason> {
     // verify validity:
     let pc = match req.cert.encoding {
         CertificateEncoding::Pem => {
@@ -42,8 +49,8 @@ pub fn verify_certificate(req: CertificateVerificationRequest) -> Result<bool> {
                 picky::x509::date::UtcDate::from(zulu_not_after);
 
             if pc.valid_not_after().lt(&p_not_after) {
-                return Err(anyhow!(
-                    "Certificate is being used after its expiration date"
+                return Ok(BoolWithReason::False(
+                    "Certificate is being used after its expiration date".to_string(),
                 ));
             }
         }
@@ -54,8 +61,8 @@ pub fn verify_certificate(req: CertificateVerificationRequest) -> Result<bool> {
 
     let now = picky::x509::date::UtcDate::now();
     if pc.valid_not_before().gt(&now) {
-        return Err(anyhow!(
-            "Certificate is being used before its validity date"
+        return Ok(BoolWithReason::False(
+            "Certificate is being used before its validity date".to_string(),
         ));
     }
 
@@ -64,10 +71,14 @@ pub fn verify_certificate(req: CertificateVerificationRequest) -> Result<bool> {
         let mut certs = vec![];
         certs.append(&mut certch);
         let cert_pool = CertificatePool::from_certificates(&certs)?;
-        cert_pool.verify(&pc)?
+        if !cert_pool.verify(&pc) {
+            return Ok(BoolWithReason::False(
+                "Certificate is not trusted by the provided cert chain".to_string(),
+            ));
+        }
     }
 
-    Ok(true)
+    Ok(BoolWithReason::True)
 }
 
 impl CertificatePool {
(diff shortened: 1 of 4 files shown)


src/callback_handler/crypto.rs
```diff
@@ -274,15 +277,15 @@ iDAKBggqhkjOPQQDAgNIADBFAiEArSsdE5dDXqAU2vM3ThT8GvTnjkWhER3l9v1j
             cert_chain: Some(cert_chain),
             not_after: None,
         };
-        assert_eq!(
-            verify_certificate(req).unwrap_err().to_string(),
-            "Certificate not issued by a trusted root"
-        );
-        Ok(())
+        let _reason = "Certificate is not trusted by the provided cert chain".to_string();
```
viccuad, line 280: this here is weird; I can't instantiate `reason` inside of the `matches!()`, but if I do it outside as it is, it says it's unused.

I couldn't find a simpler way to compare the output to the enum that I want for the tests (besides making the enum implement PartialEq), so I prefer to do `_reason` here for simplicity, as these are tests.

src/callback_handler/crypto.rs
```diff
@@ -11,10 +11,17 @@ struct CertificatePool {
     intermediates: Vec<picky::x509::Cert>,
 }
 
+#[derive(Debug)]
+/// Used as return of verify_certificate()
+pub enum BoolWithReason {
+    True,
+    False(String),
+}
```
flavio, line 19: I think you can reuse the same struct that is defined inside of the Rust SDK

src/callback_handler/crypto.rs
flavio: yeah, let's stick with this approach. But please rename `_reason` to something that doesn't start with an underscore. This symbol is reserved to identify variables that are not used, but this one is actually used.

src/callback_handler/crypto.rs
```diff
@@ -345,15 +345,15 @@ iDAKBggqhkjOPQQDAgNIADBFAiEArSsdE5dDXqAU2vM3ThT8GvTnjkWhER3l9v1j
             cert_chain: None,
             not_after: Some(Utc::now().to_rfc3339()),
         };
-        assert_eq!(
-            verify_certificate(req).unwrap_err().to_string(),
-            "Certificate is being used after its expiration date"
-        );
-        Ok(())
+        let _reason = "Certificate is being used after its expiration date".to_string();
```
flavio, line 348: same here about the `_reason` name

src/callback_handler/crypto.rs
```diff
@@ -363,10 +363,10 @@ iDAKBggqhkjOPQQDAgNIADBFAiEArSsdE5dDXqAU2vM3ThT8GvTnjkWhER3l9v1j
             cert_chain: None,
             not_after: None,
         };
-        assert_eq!(
-            verify_certificate(req).unwrap_err().to_string(),
-            "Certificate is being used before its validity date"
-        );
-        Ok(())
+        let _reason = "Certificate is being used before its validity date".to_string();
```
flavio, line 366: same here

src/runtimes/wapc.rs
```diff
@@ -129,8 +129,20 @@ pub(crate) fn host_callback(
...
+                        Ok(b) => match b {
+                            BoolWithReason::True => CertificateVerificationResponse {
+                                trusted: true,
+                                reason: "".to_string(),
+                            },
+                            BoolWithReason::False(reason) => CertificateVerificationResponse {
+                                trusted: false,
+                                reason,
+                            },
+                        },
```
flavio, line 142: nitpicking, this could be the implementation of `TryFrom`

flavio (commented):
overall ok, I left some minor comments

src/callback_handler/crypto.rs
viccuad: that's the thing, I do need the underscore, and it's being used (as proof by the tests passing by checking the correct strings). If I don't use the underscore, I will get warnings on all the tests.

src/runtimes/wapc.rs
viccuad: true! done in https://github.com/kubewarden/policy-evaluator/pull/214/commits/d21868bc1b20db0dc2f763c24f5e39e549aa6b01.

src/callback_handler/crypto.rs
viccuad: Better, yes. Done in https://github.com/kubewarden/policy-evaluator/pull/214/commits/d21868bc1b20db0dc2f763c24f5e39e549aa6b01.

src/callback_handler/crypto.rs
flavio: oh, that's super weird! Ignore my comment then. Maybe add a simple inline comment about that...

flavio (approved):

src/runtimes/wapc.rs
viccuad: The implementation of the From ends being [for CertificateVerificationResponse, here](https://github.com/kubewarden/policy-sdk-rust/pull/72/commits/a52b07c69023ba7f6f960b39a7076e9bfcd2fcbe).

viccuad:
rebased consuming the tagged version of policy-rust-sdk.

PR #224, opened 2022-12-19
feat: timeout protection

flavio:
This is quite a big change. It's required to allow consumption of the [wasmtime epoch interruption](https://docs.rs/wasmtime/3.0.1/wasmtime/struct.Config.html#method.epoch_interruption) capability. Which we leverage to allow policies to be stopped after they have exceeded a given run time.

This is required to fix https://github.com/kubewarden/policy-server/issues/254

While working on this code the following changes have been done:

- feat!: burrego - epoch deadline feature + builder: thinking about Policy Server, each worker has just one engine, which is then used to run both waPC and burrego policies. Once the engine has the epoch deadline feature turned on, some operations have to be performed to ensure the policy can be executed (add some [epoch ticks](https://docs.rs/wasmtime/3.0.1/wasmtime/struct.Store.html#method.set_epoch_deadline) to the `wasmtime::Store`). Because of that, Burrego had to be "made aware" of epoch interruption. I took advantage of this chance to do some cleanup and introduce a builder helper for the `burrego::Evaluator`
- feat!: expose epoch deadline feature: at a higher level, PolicyEvaluator now exposes epoch deadline configuration. The builder then handles the enablement of this feature both inside of waPC and burrego
- makefile: add check target: a small change done to make the development experience smoother
- Add envrc file: similar to the previous point
- wapc policy: reset evaluator after epoch interruption: while testing the feature, I realized that once a waPC guest is interrupted by the epoch deadline feature, the wasm module might be left in a broken state. See the long explanation about that inside of the code. With this commit, the waPC-based PolicyEvaluator is reset after an epoch interruption happens
- refactor![burrego]: get rid of anyhow: this is a huge refactor inside of burrego. Now the public - and internal - APIs of burrego have dedicated typed errors. This is required for later
- feat! [burrego]:introduce ExecutionDeadlineExceeded errors: catch the epoch interrupt errors raised by Wasmtime and convert them into a dedicated burrego error
- [burrego] tag v0.3.0 release: no more changes are required to burrego, doing a minor version bump
- feat: burrego evaluator - perform reset after epoch interrupt: now that burrego has typed errors, ensure that the burego evaluator is reset after a wasmtime epoch interruption error is raised


flavio:
@ereslibre can you take a look at this PR if you have some spare time?

jvanz:
@flavio is this a breaking change? I saw in the first comment bunch of `feat!`. If so, we should tag this as `kind/breaking-change`. Just to ensure that when we added the release-drafter CI the change will be properly considered.

crates/burrego/src/evaluator.rs
```diff
@@ -1,55 +1,112 @@
...
-    #[allow(dead_code)]
-    linker: Linker<Option<StackHelper>>,
+    module: Module,
     store: Store<Option<StackHelper>>,
     instance: Instance,
     memory: Memory,
     policy: Policy,
+    host_callbacks: HostCallbacks,
+    /// used to tune the [epoch
+    /// interruption](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.epoch_interruption)
```
jvanz, line 38: I have a question about this documentation formatting. Markdown link syntax work some where when reading a source code file?

src/runtimes/wapc.rs
```diff
@@ -229,6 +236,84 @@ fn send_request_and_wait_for_response(
...
+
+        Ok(Self {
+            engine,
+            module,
+            epoch_deadlines,
+            wapc_host,
+        })
+    }
+
+    /// Provision a new wapc_host. Useful for starting from a clean slate
```
jvanz, line 262:
```suggestion
    /// Provision a new wapc_host. Useful for starting from a clean state
```

jvanz (approved):

crates/burrego/src/evaluator.rs
```diff
@@ -169,12 +169,18 @@ impl Evaluator {
             data = serde_json::to_string(&data)?.as_str(),
             "setting policy data"
         );
+        if let Some(deadline) = self.epoch_deadline {
+            self.store.set_epoch_deadline(deadline);
```
viccuad: A bit clunky to need to set the deadline for each wasm evaluation separately (in this case the json parse fun). I fear I would forget some at some point. We could pass down `Option<Deadline>` and call `store.set_epoch_deadline()` inside of `policy.set_data()`, `policy.evaluate()` etc.

src/runtimes/wapc.rs
```diff
@@ -284,13 +369,57 @@ impl<'a> Runtime<'a> {
             }
             Err(e) => {
                 error!(error = e.to_string().as_str(), "waPC communication error");
+                if e.to_string()
+                    .as_str()
+                    .contains(WAPC_EPOCH_INTERRUPTION_ERR_MSG)
+                {
+                    // TL;DR: after code execution is interrupted because of an
```
viccuad, line 376: questions:

we can't use [`epoch_deadline_async_yield_and_update()`](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.epoch_deadline_async_yield_and_update) because we aren't async (we spawn different threads and each has a synchronous wasmtime::Engine). Correct?

We can't use [`epoch_deadline_callback()`](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.epoch_deadline_callback) to yield to other policy and bump the deadline, because we only have 1 policy per wasmtime::Engine, each in their own threads (good for containerization). The policy must complete evaluation before deadline, there's no yield and resume. Correct?

src/runtimes/wapc.rs
```diff
@@ -14,20 +21,20 @@ use crate::cluster_context::ClusterContext;
...
     pub(crate) static ref WAPC_POLICY_MAPPING: RwLock<HashMap<u64, Policy>> =
         RwLock::new(HashMap::with_capacity(64));
 }
 
+/// Error message returned by wasmtime_provider when the guest execution
+/// is interrupted because of epoch deadline is exceeded.
+///
+/// Unfortunately, wasmtime_provider doesn't return a typed error, hence we have
+/// to look for this text
+const WAPC_EPOCH_INTERRUPTION_ERR_MSG: &str = "guest code interrupted, execution deadline exceeded";
```
viccuad, line 36: should we add a unit test that triggers this error, and checks that the string keeps being this one? I'm scared it subtly changes in future versions of wasmtime_provider and we just skip it.

src/runtimes/burrego.rs
```diff
@@ -160,9 +160,17 @@ impl<'a> Runtime<'a> {
             }
             Err(err) => {
                 error!(
-                    error = err.to_string().as_str(),
+                    error = ?err,
```
viccuad, line 163: I don't understand this casting. Does this mean [that err is optional](https://doc.rust-lang.org/reference/macros-by-example.html#repetitions)?

viccuad (approved):
LGTM, amazing PR!

I would like a `WAPC_EPOCH_INTERRUPTION_ERR_MSG` unit test. Apart from that, I have some questions just for my benefit.

flavio:
> @flavio is this a breaking change? I saw in the first comment bunch of `feat!`. If so, we should tag this as `kind/breaking-change`. Just to ensure that when we added the release-drafter CI the change will be properly considered.

The public API of burrego has been changed in a breaking way.

The same does not apply to policy-evaluator

crates/burrego/src/evaluator.rs
flavio: I think it depends on the IDE being used by the developer. However it works when generating the html documentation via `cargo doc`.

src/runtimes/burrego.rs
flavio: This is not a casting. This is the special syntax used by the `tracing` macros. The question mark means: use the display output of the object (like doing `{:?}` formatting)

flavio: You can see more details [here](https://docs.rs/tracing/0.1.37/tracing/#recording-fields)

src/runtimes/wapc.rs
flavio: This is an [English way of saying](https://www.wordnik.com/words/clean%20slate) that means:

* > noun by extension a fresh start
* >  an opportunity to start over without prejudice

flavio: I can change it to the more formal way you proposed

src/runtimes/wapc.rs
flavio: > We don't want yield and resume of policy evaluation, because that would block the k8s resource in the admissioncontroller anyways. Correct?

We don't want to resume because we don't know if the code will ever complete its execution. After a certain amount of time the Kubernetes API server will declare the response to be a timeout. IMHO I think it's useless to keep giving control back to the wasm code hoping it completes its execution, just to end up not having anybody interested in the response (because of the timeout).

> we can't use epoch_deadline_async_yield_and_update() anyways because we aren't async (we spawn different threads in the worker pool and each has a synchronous wasmtime::Engine). Correct?

True, this is can be used only when the async feature is enabled inside of wasmtime and all the interactions with wasmtime are done via the async methods. This would require quite a big set of changes to waPC. This is something I want to contribute, but not right now.

However, I don't want to give control back to the wasm guest once the epoch interruption took place, see above.

> We can't use epoch_deadline_callback() to yield to other policy and bump the deadline, because we only have 1 policy per wasmtime::Engine, each in their own threads (good for containerization). Correct?

We could use the [Store::epoch_deadline_callback](https://docs.rs/wasmtime/4.0.0/wasmtime/struct.Store.html#method.epoch_deadline_callback) to allow the wasm code to run longer, all without having to enable async support inside of wasmtime. However, I don't want to give control back to the wasm guest once the epoch interruption took place, see above.

However, if we were to give control back, we would still need to abort the wasm guest code after some time because otherwise we would end up having one of our worker threads stuck evaluating the same request.
Other similar requests could be made against policy server, leading to all the workers being stuck in an endless evalution, leading to DOS attack against policy server. We know that would have bad effects on the Kubernetes API server and the cluster itself.

viccuad: Thanks for the detailed answers, resolving!

src/runtimes/burrego.rs
viccuad: doh, thanks!

crates/burrego/src/evaluator.rs
flavio: Yes, this got refactored by using a macro. Luckily this is something internal, that the end user doesn't have to mess with.

Can I mark this conversation as resolved?

src/runtimes/wapc.rs
flavio: This is covered by the e2e tests added to policy-server, but yes... this is something we should cover with a quick test also there. I'm going to fix that

flavio:
@viccuad, @jvanz: I think this is ready for one last round of reviews. I just rebased the code against `main` and implemented the unit test required by @viccuad

src/runtimes/wapc.rs
jvanz: Oh, that's nice! Cool new expression learnt. Thanks! No need for change.

crates/burrego/src/evaluator.rs
viccuad: yep, totally, resolving!

jvanz (approved):
LGTM. The WebAssembly code is very interesting.

PR #246, opened 2023-02-20
Add `ArtifactHubPkg{}`, `ArtifactHubPkg.from_metadata()`

viccuad:
## Description

Relates to https://github.com/kubewarden/kwctl/issues/418

Following https://github.com/kubewarden/rfc/blob/main/rfc/0014-metadata.md,
Add `ArtifactHubPkg{}`.
Add `from_metadata()`. Since there's no way to iterate through 

Second try (after typing annotations as simple Strings).
For annotations, I have considered [strum](https://crates.io/crates/strum), [EnumSet](https://docs.rs/enumset/latest/enumset/) (with [serde_with](https://crates.io/crates/serde_with)), yet EnumSet variants don't accept fields.


src/policy_artifacthub.rs
```diff
@@ -0,0 +1,263 @@
...
+    #[serde(rename = "homeURL", skip_serializing_if = "Option::is_none")]
+    pub home_url: Option<Url>, // https://github.com/kubewarden/verify-image-signatures
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub containers_images: Option<Vec<ContainerImage>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub keywords: Option<Vec<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub links: Option<Vec<Link>>,
+
+    // #[serde(skip_serializing_if = "Option::is_none")]
```
viccuad, line 32: I'm undecided on if to use `metadata.yml`'s `"io.kubewarden.policy.author"` to construct one or both `artifacthubpkg.yml`'s `provider` and `maintainers`. Hardcoding all `provider: kubewarden` may be ok, but having no distinction between policies owned by the Kubewarden team and 3rd parties is weird.

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub url: Url,
+}
+
+fn default_recommendation() -> Url {
+    Url::parse("https://artifacthub.io/packages/helm/kubewarden/kubewarden-controller").unwrap()
+}
+
+// #[derive(Default, Deserialize, Serialize, Debug, Clone)]
+// #[serde(rename_all = "camelCase")]
+// pub struct Maintainer {
```
viccuad, line 97: This should conform `address-list` following https://www.rfc-editor.org/rfc/rfc2822#section-3.
Maybe implementable with https://docs.rs/mail-parser/latest/mail_parser/struct.Addr.html, but doubt it can be deserialized without writing a deserialize.

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,325 @@
...
+                url: default_recommendation(),
+            }],
+            annotations: HashSet::from([
+                Annotation::KubewardenMutation(metadata.mutating),
+                Annotation::KubewardenContextAware(metadata.context_aware),
+                Annotation::KubewardenRules(
+                    serde_json::to_string(&metadata.rules).expect("policy rules key is malformed"),
+                ),
+            ]),
+            // the following required fields will be filled later on:
```
viccuad, line 169: this is the incredibly ugly part.
it could be improved by setting those required fields right here and then. For example, for the `artifacthubpkg.name` field:

```rust
            name: {
                let mut missing_title = false;
                for a in metadata_annots { // could be 0 loops, but we checked is not empty above..
                   if let policy_metadata::Annotation::PolicyTitle(ref title) = a {
                       // found
                       title
                    } else {
                        // not found
                    }
                    if missing_title {
                        return Err(anyhow!(
                            "policy metadata must specify {} in annotations",
                            "\"io.kubewarden.policy.title\"",
                        ));
                    }
                }
            },
```

Yet something like this is not easily parametrized to transform as function. As we need to do this for `String` and `Url` (`Link`)

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,325 @@
...
+                "policy metadata must specify {} in annotations",
+                "\"io.kubewarden.policy.description\"",
+            ));
+        }
+        if artifacthubpkg.links.is_none() {
+            return Err(anyhow!(
+                "policy metadata must specify {} in annotations",
+                "\"io.kubewarden.policy.source\"",
+            ));
+        }
```
flavio, line 291: This is ok, but we could maybe use [`validator`](https://crates.io/crates/validator) for that. This is a crate we already use inside of policy-evaluator

src/policy_artifacthub.rs
flavio: That, or the loop you wrote below. I've just seen that

src/policy_artifacthub.rs
flavio: Right now `io.kubewarden.policy.author` is a string that can hold multiple mail addresses. I would use that as input data to build the `maintainers` key, which is the only one that 1) allows multiple people to be used 2) has the concept of name/email.

I would just ignore `provider` and even leave it out of our struct

src/policy_artifacthub.rs
flavio: Why do we need a custom deserializer?

What I would do is:

* Look for the `io.kubewarden.policy.author` field inside of `metadata.yml`
* Split it by `,`
* Iterate over each entry and try to parse that with the crate you linked
* We should just consider one email format: `Homer Simpson <homer@simpsons.com>`
* If parsing fails, exit with an error and tell what is the email format we support

src/policy_metadata.rs
```diff
@@ -141,6 +142,51 @@ const fn _default_true() -> bool {
     true
 }
 
+#[derive(Deserialize, Serialize, Debug, Clone, Validate)]
+#[serde(rename_all = "camelCase")]
+#[validate(schema(function = "validate_metadata_v2", skip_on_field_errors = false))]
+pub struct MetadataV2 {
```
flavio, line 148: I'm lost, why do we need a Metadata `V2`?

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,325 @@
...
+            license: Default::default(),
+            home_url: Default::default(),
+            containers_images: Default::default(),
+            keywords: Default::default(),
+            links: Default::default(),
+            provider: Default::default(),
+            recommendations: Default::default(),
+            annotations: Default::default(),
+            readme: Default::default(),
+        }
```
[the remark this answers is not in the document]
viccuad: It can't be used here :/, it applies to `Self`, and makes the assignation recursive: https://internals.rust-lang.org/t/pre-pre-rfc-the-default-default-shorthand-is-misleading/15917.

src/policy_artifacthub.rs
viccuad: Built maintainers out of `io.kubewarden.policy.author`.
Hardcoded `provider` to `kubewarden` as per RFC.

src/policy_artifacthub.rs
viccuad: Done it by manually parsing name and email.

src/policy_artifacthub.rs
viccuad: Resolving, as this is irrelevant now.

src/policy_artifacthub.rs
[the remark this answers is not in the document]
viccuad: Resolving, irrelevant after refactor.

src/policy_artifacthub.rs
viccuad: I've made the checks in separate functions, by pattern matching instead of validate functions. Please tell me what you think about it.

src/policy_metadata.rs
viccuad: We don't need it indeed! resolving, irrelevant after the refactor.

viccuad:
Completed refactor.
More tests should be added, particularly about specific errors on custom parsing (in `parse_keywords()`, `parse_links()`, `parse_maintainers()`, `parse_annotations()`).

viccuad:
Also implemented the changes to the RFC done by https://github.com/kubewarden/rfc/pull/18.

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,325 @@
...
+///  - name: policy
+///    url: https://github.com/kubewarden/verify-image-signatures/releases/download/v0.2.1/policy.wasm
+///  - name: source
+///    url: https://github.com/kubewarden/verify-image-signatures
+#[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct Link {
+    pub name: String, // either "policy", or "source"
+    pub url: Url,
+}
```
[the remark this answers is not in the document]
viccuad: Done!

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,325 @@
...
+            readme: Default::default(),
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct ContainerImage {
+    pub name: String,  // always "policy"
+    pub image: String, // not URL, e.g: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.1
```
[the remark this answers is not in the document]
viccuad: Done!

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,325 @@
...
+        if metadata_annots.is_empty() {
+            return Err(anyhow!(
+                "no annotations in policy metadata. policy metadata must specify annotations"
+            ));
+        }
+
+        // build minimal struct with required fields, besides those derived from
+        // metadata.annotations:
+        let mut artifacthubpkg = ArtifactHubPkg {
+            version: Version::parse(version).expect("policy version must be in semver"),
```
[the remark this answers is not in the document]
viccuad: Done!

jvanz (approved):
LGTM

viccuad:
Added missing tests, ready for merge.

src/constants.rs
```diff
@@ -1,10 +1,24 @@
...
 pub const KUBEWARDEN_ANNOTATION_POLICY_URL: &str = "io.kubewarden.policy.url";
+pub const KUBEWARDEN_ANNOTATION_POLICY_OCIURL: &str = "io.kubewarden.policy.ociUrl";
 pub const KUBEWARDEN_ANNOTATION_POLICY_SOURCE: &str = "io.kubewarden.policy.source";
 pub const KUBEWARDEN_ANNOTATION_POLICY_LICENSE: &str = "io.kubewarden.policy.license";
 pub const KUBEWARDEN_ANNOTATION_POLICY_USAGE: &str = "io.kubewarden.policy.usage";
 
+pub const KUBEWARDEN_ANNOTATION_ARTIFACTHUB_RESOURCES: &str = "io.artifacthub.resources";
+pub const KUBEWARDEN_ANNOTATION_ARTIFACTHUB_DISPLAYNAME: &str = "io.artifacthub.displayName";
+pub const KUBEWARDEN_ANNOTATION_ARTIFACTHUB_KEYWORDS: &str = "io.artifacthub.keywords";
+pub const KUBEWARDEN_ANNOTATION_RANCHER_HIDDENUI: &str = "io.rancher.hidden-ui";
```
viccuad, line 15: I wanted to rename to:

```suggestion
pub const KUBEWARDEN_ANNOTATION_RANCHER_HIDDENUI: &str = "io.kubewarden.hidden-ui";
```

Together with
```
pub const ARTIFACTHUB_ANNOTATION_RANCHER_HIDDENUI: &str = "kubewarden/hidden-ui";
```

In preparation for a standalone UI. This means changes in UI, policies, RFC, but I wonder if it's worth it.

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+        let arthub = ArtifactHubPkg::from_metadata(
+            &mock_metadata_with_minimum_required(),
+            "not-semver",
+            OffsetDateTime::UNIX_EPOCH,
+            None,
+        );
+        assert_eq!(
+            arthub.unwrap_err().to_string(),
+            "policy version must be in semver: unexpected character 'n' while parsing major version number"
+        );
```
flavio, line 627: It's not your fault, there's no other way to do that... But I think we should really get rid of `anyhow` and use `thiserror`.

I see two possibilities:

* We leave the code as it is, we merge this PR and later we do a big PR that kills `anyhow::Result`
* You change this PR: you introduce a first version of our `Error` type, with just the errors that the code of this PR introduces. You refactor the tests to make use of this Error. Later we do another PR that extends the custom error you introduced and just removes anyhow

flavio (commented):
LGTM, I just left some minor comments

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+                }
+            }
+            for person in csv.iter() {
+                let vec_person = person.split('<').collect::<Vec<&str>>();
+                if vec_person.len() != 2 {
+                    return Err(anyhow!(
+                        "annotation \"{}\" in policy metadata is malformed, elements should be in \"name <email address>\" format",
+                        KUBEWARDEN_ANNOTATION_POLICY_AUTHOR
+                    ));
+                };
```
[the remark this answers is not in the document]
viccuad: Sadly, no.
I've tried to parse the `name-addr`, so, `Tux <tux@example.com>` according to https://www.rfc-editor.org/rfc/rfc5322#section-3.4) with:
   https://docs.rs/mail-parser/0.8.2/mail_parser/struct.Addr.html
and
  https://github.com/stalwartlabs/smtp-proto/blob/main/src/request/parser.rs#L401
  
 But without luck to be honest.
 I can keep looking a bit more, but I'm happy with the current approach: naively separate `display-name` and `angle-addr`, check the mail address inside `addr-spec`.

src/constants.rs
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/b5a525f5a705fc045d3966e948b29e1f745e2fe5,
also https://github.com/kubewarden/rfc/pull/19.

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+
+impl ArtifactHubPkg {
+    pub fn from_metadata(
+        metadata: &Metadata,
+        version: &str,
+        created_at: OffsetDateTime,
+        questions: Option<&str>,
+    ) -> Result<Self> {
+        // validate inputs
+        //
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/f0c3f39a22788e7b9c78d9cc02d3edb4b8a53805

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+        }
+        let metadata_annots = metadata.annotations.as_ref().unwrap();
+        if metadata_annots.is_empty() {
+            return Err(anyhow!(
+                "no annotations in policy metadata. policy metadata must specify annotations"
+            ));
+        }
+        let semver_version = Version::parse(version)
+            .map_err(|e| anyhow!("policy version must be in semver: {}", e))?;
+        if questions.is_some() && questions.unwrap().is_empty() {
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/cf93ea3c65ce796f394c89001b6b522a77ae7e11.
I'm ok leaving passing questions as a `Some(s)` because it's optional, so then `parse_annotations()` can see if they are there, and add the corresponding annotation for them or not.

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+                .get(KUBEWARDEN_ANNOTATION_POLICY_LICENSE)
+                .cloned(),
+            home_url,
+            containers_images,
+            keywords,
+            links,
+            maintainers,
+            provider: Provider {
+                name: ConstProvider::Kubewarden,
+            },
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/ce25bc36e998c05f4f6a67b02912f8ce61793b77

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+fn parse_name(metadata_annots: &HashMap<String, String>) -> Result<String> {
+    match metadata_annots.get(KUBEWARDEN_ANNOTATION_POLICY_TITLE) {
+        Some(s) => Ok(s.to_string()),
+        None => {
+            return Err(anyhow!(
+                "policy metadata must specify \"{}\" in annotations",
+                KUBEWARDEN_ANNOTATION_POLICY_TITLE,
+            ))
+        }
+    }
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/3a0683c0ee299366946c9a4d8b867c7a1919ed9a

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+fn parse_display_name(metadata_annots: &HashMap<String, String>) -> Result<String> {
+    match metadata_annots.get(KUBEWARDEN_ANNOTATION_ARTIFACTHUB_DISPLAYNAME) {
+        Some(s) => Ok(s.to_string()),
+        None => {
+            return Err(anyhow!(
+                "policy metadata must specify \"{}\" in annotations",
+                KUBEWARDEN_ANNOTATION_ARTIFACTHUB_DISPLAYNAME,
+            ))
+        }
+    }
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/3a0683c0ee299366946c9a4d8b867c7a1919ed9a

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+                .map(str::to_string)
+                .collect::<Vec<String>>();
+            for word in csv.iter() {
+                if word.is_empty() {
+                    return Err(anyhow!(
+                        "annotation \"{}\" in policy metadata is malformed, must be csv values",
+                        KUBEWARDEN_ANNOTATION_ARTIFACTHUB_KEYWORDS
+                    ));
+                }
+            }
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/3a0683c0ee299366946c9a4d8b867c7a1919ed9a

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,923 @@
...
+                .map(str::to_string)
+                .collect::<Vec<String>>();
+            for word in csv.iter() {
+                if word.is_empty() {
+                    return Err(anyhow!(
+                        "annotation \"{}\" in policy metadata is malformed, must be csv values",
+                        KUBEWARDEN_ANNOTATION_POLICY_AUTHOR
+                    ));
+                }
+            }
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/3a0683c0ee299366946c9a4d8b867c7a1919ed9a

src/constants.rs
viccuad: @jordojordo would it be possible to honour https://github.com/kubewarden/rfc/pull/19 in the UI?

jordojordo: @viccuad This has been accomplished with [this pr](https://github.com/kubewarden/ui/pull/264)

src/policy_artifacthub.rs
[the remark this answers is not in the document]
flavio: I found a way to use `mail_parser`:

```rust
fn main() {
    let maintainers = "Homer Simpson <homer@simpson.com>, Marge Simpson <marge@simpson>";
    let to = format!("To: {maintainers}");

    let msg = mail_parser::Message::parse(to.as_bytes()).unwrap();

    let addr = msg.to();

    println!("Hello: {addr:?}");
}
```

The program produces this output:

```console
Hello: AddressList([Addr { name: Some("Homer Simpson"), address: Some("homer@simpson.com") }, Addr { name: Some("Marge Simpson"), address: Some("marge@simpson") }])
```

src/policy_artifacthub.rs
flavio: Let's introduce `thiserror` in a single shot, later on

flavio (approved):
LGTM. Maybe the only thing to change is the usage of `mail_parser`, see my snippet. I think it would be nice to leverage this library and get rid of our code. However I don't consider that as blocking

src/policy_artifacthub.rs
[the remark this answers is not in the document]
viccuad: still, that parser returns an Option, not a Result. It can contain a full mail message if the string is crafted on purpose.
Note that we still need the second lib, `email_address`, to parse the mail address itself, as `mail_parser` doesn't do it..
I keep looking.

src/constants.rs
viccuad: Thanks @jordojordo !

src/policy_artifacthub.rs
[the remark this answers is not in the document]
viccuad: Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/9338445f091c3949a037e262fe5d2ed070ca2254.

src/policy_artifacthub.rs
```diff
@@ -0,0 +1,933 @@
...
+
+    #[error("annotation \"{0}\" in policy metadata is malformed, must be csv values of \"name <email>\"")]
+    MalformedCSVEmail(String),
+
+    #[error("annotation \"{annot:?}\" in policy metadata must be a well formed URL: {error:?}")]
+    MalformedEmail { annot: String, error: String },
+
+    #[error("annotation \"{0}\" in policy metadata is malformed, must be a string \"true\" or \"false\"")]
+    MalformedBoolString(String),
+}
```
flavio, line 54: nitpicking: I would move that to a new `errors.rs` file

flavio (approved):
LGTM

flavio (approved):
LGTM

flavio:
@viccuad the code LGTM, but the tests are now broken. Can you look into that please?

viccuad:
Finally ready to merge.

src/policy_artifacthub.rs
flavio: Marking as unresolved, now that you have introduced `thiserror` the unit tests should look for the error type, not for its string representation

flavio (changes requested):
Sorry for the nitpicking, but the tests have to be updated now that we have `thiserror`

viccuad:
Indeed, forgot to do that. Done!

src/policy_artifacthub.rs
```diff
@@ -577,8 +571,10 @@ mod tests {
             None,
         );
         assert_eq!(
-            arthub.unwrap_err().to_string(),
-            "policy version must be in semver: unexpected character 'n' while parsing major version number"
+            arthub.unwrap_err(),
+            ArtifactHubError::NoSemverVersion(String::from(
```
flavio, line 575: I'm fine with it, but there's still the chance to have the test break because we change something inside of the error string.

I would do something like that:

```rust
assert!(matches!(arthub.unwrap_err(), ArtifactHubError::NoSemverVersion))
```

I think checking for the error type is good enough.

There are also other error checks below that look after specific strings, I don't know if it would make sense to change them too

flavio (approved):
LGTM, I left one minor comment that can be ignored

src/policy_artifacthub.rs
viccuad: True, better. Done in https://github.com/kubewarden/policy-evaluator/pull/246/commits/4338dd4866ab508a0b90d3180f1f7aaa38f7420f.

flavio (approved):
Nice, thanks for all the fixes

viccuad:
Tagged as 0.7.0 in main.

PR #247, opened 2023-02-21
Context aware - foundation

flavio:
> **Note:** this PR is made against a feature branch, not `main`

This is the initial work required to get the new context aware implementation ready.

Once merged, this allows policies to fetch any kind of Kubernetes resource from a cluster.
No security check is done yet, meaning a policy could fetch basically everything. This will be fixed with a future PR made against this feature branch.

These are the commits done inside of this PR, each one aims to be self explanatory:

- refactor: split and DRY code
- Update policy metadata
- feat: introduce new context aware callbacks
- chore(deps): use latest version of the SDK


src/callback_handler/mod.rs
```diff
@@ -107,6 +30,28 @@ pub struct CallbackHandler {
...
+                debug!(
+                    value = ?$log_value,
+                    cached = response.was_cached,
+                    $log_msg,
+                );
+                let payload = serde_json::to_vec(&response.value)
+                    .map_err(|e| anyhow!("error serializing payload: {e:?}"))?;
+                Ok(CallbackResponse { payload })
+            })
+            .and_then(|r| r);
```
viccuad, line 49: this waits until the response Future has succeeded and returns it. If not, we issue the warn, correct?
We are tracing it as warn because that's what we did. Should it be error?

flavio: It waits for the future to complete. The response returns a `Result<something>`. If the response is `Ok` we enter the `map(|response|)` block, which converts the response object into a json string.

However, the transformation can fail too. Hence the `map` block returns a `Result<Result<Vec<u8>>>`. The `and_then` block is executed **only if** the original response was `Ok`, it's purpose is to flatten the `Result<Result<>>` to be a `Result<Vec<u8>>`.

Long story short: this is convoluted, but no error is lost during the process. There's a rust nightly feature called [`Result::flatten`](https://doc.rust-lang.org/std/result/enum.Result.html#method.flatten) which would be more explicit, but this works in the same way

src/callback_handler/kubernetes.rs
```diff
@@ -0,0 +1,201 @@
...
+                            version: resource.version.unwrap_or_default(),
+                            api_version: api_version.to_string(),
+                            kind: kind.to_string(),
+                            plural: resource.name,
+                        },
+                        namespaced: resource.namespaced,
+                    })
+                    .clone()
+            }
+            hash_map::Entry::Occupied(entry) => entry.get().clone(),
```
viccuad, line 85: IIUC, we cache twice; once here in the client's `kube_resources`, and another time in the `pub list_resources_*()` (he `list_resources_*()` caches don't take into account the client).
Is that ok? Wouldn't it be ok with one of both?

viccuad (approved):
LGTM! Just one question

src/callback_handler/mod.rs
viccuad: I see it now. Thanks!

src/callback_handler/kubernetes.rs
flavio: It's fine to not take the client into account when building this internal cache of `KubeResource` objects. We are always connected to the same Kubernetes cluster.

The only improvement I could make would be to put this HashMap behind a RwLock. Policy Server has multiple workers running policies, but all of them end up invoking the same host capabilities handler. There's the risk of having two concurrent inserts.

I'll fix that

flavio: Done with 9d67a0b

flavio:
@viccuad can you take another look please?

src/callback_handler/kubernetes.rs
viccuad: Good catch, thanks!

PR #249, opened 2023-02-21
Limit the k8s resources that can be requested by a policy

flavio:
Limit the k8s resources that can be requested by a policy

Starting from this commit each policy has a list of Kubernetes resources that is allowed to access.

Any attempt to access a non allowed resource is going to be blocked and reported via the system logs.


viccuad (approved):
LTGM!

PR #256, opened 2023-03-07
context aware: fix artifact hub metadata

flavio:
Update artifact hub metadata to reflect the changes done to context aware policies.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -17,7 +17,8 @@ pub const KUBEWARDEN_ANNOTATION_ARTIFACTHUB_HIDDENUI: &str = "io.kubewarden.hidd
 pub const KUBEWARDEN_ANNOTATION_KWCTL_VERSION: &str = "io.kubewarden.kwctl";
 
 pub const ARTIFACTHUB_ANNOTATION_KUBEWARDEN_MUTATION: &str = "kubewarden/mutation";
-pub const ARTIFACTHUB_ANNOTATION_KUBEWARDEN_CONTEXTAWARE: &str = "kubewarden/contextAware";
+pub const ARTIFACTHUB_ANNOTATION_KUBEWARDEN_CONTEXTAWARE_RESOURCES: &str =
+    "kubewarden/contextAwareResources";
 pub const ARTIFACTHUB_ANNOTATION_KUBEWARDEN_RESOURCES: &str = "kubewarden/resources";
 pub const ARTIFACTHUB_ANNOTATION_RANCHER_HIDDENUI: &str = "kubewarden/hidden-ui";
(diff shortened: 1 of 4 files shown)


jvanz (approved):
LGTM.

I'm not sure why there are error messages like this:
```
```

viccuad (changes requested):
I'm not sure this is correct.

From the [RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0015-context-aware-policies.md#policy-metadata), we start ignoring the `context_aware` bool field, and do things when `context_aware_resources` is present.

I would:
1. Decide if to set [`metadata.context_aware_resources`](https://github.com/flavio/policy-evaluator/blob/context-aware/src/policy_metadata.rs#L147) as `Option<HashSet<ContextAwareResource>>`, or  treat empty as a policy that is not context aware.
2. Start ignoring completely the legacy `metadata.context_aware`. But we are declaring `skip_on_field_errors = false`, so care needs to be taken there.
3. According to decision in 1, solve the cargo check failures in the test of artifacthub (they are looking for a field `context_aware_resources` and finding a `context_aware`).

flavio:
@viccuad you're right. I'll update the PR to have the optional hashset that you proposed. I think this information is going to be useful to the Kubewarden UI and, eventually, to ArtifactHUB itself (they could render that in a "fancy" way)

flavio:
@viccuad done, can you take another look please?

viccuad (commented):
Ok, we are going with `HashSet<ContextAwareResource>`.
IIUC, this means that we always expect `contextAwareResources` field in metadata.yml, and we always create a `kubewarden/contextAwareResources` annotation in artifacthub-pkg.yml. The contents of one go to the other.
Then:
- If it is empty, the policy is not context aware.
- If it is has some resources, the policy is context aware and wants to listen for those resources.

But this is contrary to the RFC:
> We will start to ignore this legacy flag and consider only the existence of the contextAwareResources field to determine whether a policy is context aware or not.

I'm ok with one approach or the other, but I would like to amend one or the other to be consistent.

Also, I think that the context-aware branch is still not backwards compatible with those policies that have annotation `contextAware`  (see [skip_on_errors_field false on Metadata{}](https://github.com/kubewarden/policy-evaluator/blob/72edd109f91fe19a7e9cadfc3a6922d7675f5e7a/src/policy_metadata.rs#L132)).
We can sort that out on a follow-up PR though.

flavio:
I've changed the code that creates the `artifacthub-pkg.yml` to ignore empty `contextAwareResources` at serialization time.

I've also extended the test suite to ensure that old metadata, the one featuring the now removed `contextAware` boolean flag, can still be parsed into something that makes sense. Apparently that worked out of the box :)

viccuad (approved):
LGTM!

viccuad (approved):

PR #268, opened 2023-03-29
Minimum Kubewarden version field.

### LINKED ISSUE #267: Feature Request: allow policy to define minimum version of Kubewarden required
### Is your feature request related to a problem?

Some policies might leverage host capabilities that are available only starting from a certain Kubewarden release. The policy would be "loadable" by an older version of Kubewarden, but at runtime its evaluation will not work as expected (errors will be raised by the host, leading to the rejection of the request).

### Solution you'd like

Each policy author should be allowed to specify a minimum version of the Kubewarden stack that is required by its policy.

This data is going to be added to the `metadata.yml` by the policy author.

We have to introduce a new optional attribute called that takes as value a semantic version.

### Alternatives you've considered

_No response_

### Anything else?

We have to extend `kwctl` and `policy-server` to keep this information into account. This is going to be covered by dedicated issues.

jvanz:
## Description

Adds in the Policy struct a new field to store the minimum version of the Kubewarden stack required to run the policy.

Fix #267 

## Test

```shell
make test
```

### DIFF
--- a/src/policy.rs
+++ b/src/policy.rs
@@ -161,4 +161,24 @@ mod tests {
             &requested_resource.kind
         ));
     }
+
+    #[test]
+    fn policy_fields_test() {
+        let id = "test".to_string();
+        let policy_id = Some(1);
+        let callback_channel = None;
+        let ctx_aware_resources_allow_list = Some(HashSet::new());
+
+        let policy = Policy::new(
+            id.clone(),
+            policy_id.clone(),
+            callback_channel.clone(),
+            ctx_aware_resources_allow_list.clone(),
+        )
+        .expect("cannot create policy");
+
+        assert!(policy.id == id);
+        assert!(policy.instance_id == policy_id);
+        assert!(policy.ctx_aware_resources_allow_list == ctx_aware_resources_allow_list.unwrap());
+    }
 }
--- a/src/policy_metadata.rs
+++ b/src/policy_metadata.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use kubewarden_policy_sdk::metadata::ProtocolVersion;
+use semver::Version;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use std::path::Path;
@@ -145,6 +146,8 @@ pub struct Metadata {
     #[serde(default)]
     #[validate]
     pub context_aware_resources: HashSet<ContextAwareResource>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub minimum_kubewarden_version: Option<Version>,
 }
 
 const fn _default_true() -> bool {
@@ -161,6 +164,7 @@ impl Default for Metadata {
             background_audit: true,
             execution_mode: PolicyExecutionMode::KubewardenWapc,
             context_aware_resources: HashSet::new(),
+            minimum_kubewarden_version: None,
         }
     }
 }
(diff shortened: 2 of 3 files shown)


src/policy.rs
```diff
@@ -33,6 +33,11 @@ pub struct Policy {
     /// List of ContextAwareResource the policy is granted access to.
     /// Currently, this is relevant only for waPC based policies
     pub ctx_aware_resources_allow_list: HashSet<ContextAwareResource>,
+
+    /// Minimal Kubewarden stack version required to run the policy.
+    /// This is relevant if a policy requires a host capability available after
+    /// some specific Kubewarden version.
+    pub minimal_kubewarden_version: Option<String>,
```
flavio, line 40: I would instead use a [`Version`](https://docs.rs/semver/1.0.17/semver/struct.Version.html) object to store that. I've seen this crate has also `serde` support, I think this can be used to nicely serialize/deserialize the string into this object.

The main advantage of doing that would be to immediately spot wrong input provided by the user

src/policy_evaluator_builder.rs
```diff
@@ -42,6 +42,7 @@ pub struct PolicyEvaluatorBuilder {
     wasmtime_cache: bool,
     epoch_deadlines: Option<EpochDeadlines>,
     ctx_aware_resources_allow_list: HashSet<ContextAwareResource>,
+    minimal_kubewarden_version: Option<String>,
```
flavio, line 45: I **think** we don't need this field here. This field is needed only to perform a comparison between the kubewarden stack supported by the host and the one required by the policy.
The problem is the kubewarden stack cannot be inferred by `policy-evaluator` as a library. This is something can be done only by the host (kwctl/policy-server).

Because of that, I would try to write the check on the host side, roughly:

* pull the policy
* read the policy metadata
* perform the version check -> exit with an error
* create `PolicyEvaluatorBuilder`, life goes on as usual

flavio (commented):
I left some comments. My personal tip, especially about the `PolicyEvaluatorBuilder` changes, is to start doing the changes to kwct/policy-server. Build them against the local development copy of `policy-evaluator` and then, based on what you learn there, do further changes to this PR

src/policy_metadata.rs
```diff
@@ -145,6 +146,9 @@ pub struct Metadata {
     #[serde(default)]
     #[validate]
     pub context_aware_resources: HashSet<ContextAwareResource>,
+    #[serde(default)]
```
flavio, line 149: I think this is not needed for `Option`, see above for other optional fields

src/policy_metadata.rs
```diff
@@ -631,4 +636,30 @@ mod tests {
...
+
+    #[test]
+    fn metadata_with_min_kubewarden_version() -> Result<(), ()> {
+        let metadata = Metadata {
+            protocol_version: Some(ProtocolVersion::V1),
+            annotations: None,
+            minimum_kubewarden_version: Version::parse("1.6.0").ok(),
+            ..Default::default()
+        };
+        println!("{:?}", metadata);
```
flavio, line 648: leftover

src/policy_metadata.rs
```diff
@@ -631,4 +636,30 @@ mod tests {
 
         assert!(metadata.validate().is_err());
     }
+
+    #[test]
+    fn metadata_with_min_kubewarden_version() -> Result<(), ()> {
```
flavio, line 641: I don't think this test provides much value, this is basically testing serde ability to serialize to json something... :thinking:

flavio (approved):
LGTM

PR #288, opened 2023-05-31
chore(deps): update policy-fetcher

flavio:
This is required to use latest version of sigstore

Once merged I'll tag a new release


flavio:
New version tagged

PR #306, opened 2023-07-14
feat: Expand install msg on artifacthub

viccuad:
## Description

I find the install modal on artifacthub lacking:

https://artifacthub.io/packages/kubewarden/safe-annotations/safe-annotations?modal=install

This is an attempt to improve it, as I was typing too much on my own.

### DIFF
--- a/src/policy_artifacthub.rs
+++ b/src/policy_artifacthub.rs
@@ -375,6 +375,10 @@ fn compose_install(oci_url: String) -> String {
 ```console
 kwctl pull {}
 ```
+Then, generate the policy manifest and tune it to your liking. For example:
+```console
+kwctl scaffold manifest -t ClusterAdmissionPolicy registry://{}
+```
 "#,
         oci_url
     )


src/policy_artifacthub.rs
```diff
@@ -375,6 +375,10 @@ fn compose_install(oci_url: String) -> String {
 ```console
 kwctl pull {}
 ```
+Then, generate the policy manifest and tune it to your liking. For example:
+```console
+kwctl scaffold manifest -t ClusterAdmissionPolicy registry://{}
```
viccuad, line 380: Note that here I needed to add `registry://` while for `kwctl pull` one doesn't need to.
Opened https://github.com/kubewarden/kwctl/issues/537. Maybe worth fixing it at the same time as this for simplicity.

src/policy_artifacthub.rs
```diff
@@ -375,8 +375,12 @@ fn compose_install(oci_url: String) -> String {
...
+```
 "#,
-        oci_url
+        oci_url, oci_url
```
flavio, line 383: I would change the way we build this string to make use of the new string interpolation used by latest versions of rust. I think it would make things more readable:

Do something like 

```
kwctl scaffold manifest -t ClusterAdmissionPolicy registry://{oci_url}
```

So that we can drop the line referenced by this comment

flavio (approved):

jvanz (approved):

PR #341, opened 2023-09-11
chore(deps): fix GH action dependencies

flavio:
Express all the dependencies using their full version


fabriziosestito (approved):
LGTM, version and sha double-checked

PR #357, opened 2023-10-20
feat: raw policy validation

fabriziosestito:
## Description

Change the policy evaluator and the existing runtimes to support "raw" policy requests.
Raw requests are generic requests that are not in the form of Kubernetes AdmissionRequests.
The AdmissionRequest type has been moved from the policy-server to this repo, which already contains the AdmissionResponse type.

The ValidatieRequest type has been refactored into an Enum type to support the new variant and runtimes have been changed accordingly.
The only exception is the Gatekeeper execution mode, being Gatekeeper k8s only by nature.

Note that we are still using the AdmissionResponse type as the return type of an evaluation since it is generic enough. The only difference is that we will not wrap it inside an AdmissionReview in the policy server if the user has requested a raw policy validation.

Related to: https://github.com/kubewarden/kubewarden-controller/issues/527

## Test
Since the current testing is structured this way, most of the tests are done in the policy-server e2e tests.
We could improve this by adding runtime-related integration tests inside this crate.

See: https://github.com/kubewarden/policy-evaluator/issues/358

### DIFF
--- a/src/policy_evaluator.rs
+++ b/src/policy_evaluator.rs
@@ -199,7 +199,7 @@ mod tests {
 
         for (mode_str, expected) in &test_data {
             let actual: std::result::Result<PolicyExecutionMode, serde_json::Error> =
-                serde_json::from_str(&mode_str);
+                serde_json::from_str(mode_str);
             assert_eq!(expected, &actual.unwrap());
         }
 
--- a/src/policy_evaluator.rs
+++ b/src/policy_evaluator.rs
@@ -33,6 +33,8 @@ impl fmt::Display for PolicyExecutionMode {
     }
 }
 
+/// A validation request that can be sent to a policy evaluator.
+/// It can be either a raw JSON object, or a Kubernetes AdmissionRequest.
 #[derive(Clone, Debug, Serialize)]
 #[serde(untagged)]
 pub enum ValidateRequest {
--- a/src/policy_evaluator.rs
+++ b/src/policy_evaluator.rs
@@ -5,6 +5,7 @@ use serde::Serialize;
 use serde_json::value;
 use std::{convert::TryFrom, fmt};
 
+use crate::admission_request::AdmissionRequest;
 use crate::admission_response::AdmissionResponse;
 use crate::policy::Policy;
 use crate::runtimes::burrego::Runtime as BurregoRuntime;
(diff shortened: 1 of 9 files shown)


viccuad (approved):
LGTM. Looking forward to the RFC and the possibility of different endpoints in policy-server!

Given that this is a POC, is it ok with merging against main? (or at least, should we wait for the test infra work?)

src/policy_evaluator.rs
```diff
@@ -32,19 +33,26 @@ impl fmt::Display for PolicyExecutionMode {
...
-            uid
-        } else {
-            ""
+        match self {
+            ValidateRequest::Raw(raw_req) => {
+                if let Some(uid) = raw_req.get("uid").and_then(value::Value::as_str) {
+                    uid
+                } else {
+                    ""
+                }
```
flavio, line 53: IMHO this looks better:
```suggestion
            ValidateRequest::Raw(raw_req) => raw_req
                .get("uid")
                .and_then(value::Value::as_str)
                .unwrap_or_default(),
```

src/runtimes/burrego.rs
```diff
@@ -44,16 +44,35 @@ impl<'a> Runtime<'a> {
...
-                        "kind": "AdmissionReview",
-                        "request": &request.0,
-                    }),
-                    json!(settings),
-                )
+                match request {
+                    ValidateRequest::AdmissionRequest(adm_req) => (
+                        json!({
+                            "apiVersion": "admission.k8s.io/v1",
+                            "kind": "AdmissionReview",
```
flavio, line 51: I think you said we can omit these 2 lines. Can you drop them if that's true? That would reduce the amount of code

src/runtimes/wapc.rs
```diff
@@ -513,14 +513,18 @@ impl<'a> Runtime<'a> {
...
+            ValidateRequest::Raw(_) => Some(&req_json_value),
+            ValidateRequest::AdmissionRequest(_) => req_json_value.get("object"),
+        };
 
         let validate_params = json!({
             "request": request,
             "settings": settings,
         });
 
+        //NOTE: object is null for DELETE operations
```
flavio, line 527: Shouldn't we move this comment to line 518?

src/runtimes/wasi_cli/runtime.rs
```diff
@@ -119,11 +119,19 @@ impl<'a> Runtime<'a> {
                     )
                 }
                 match serde_json::from_slice::<PolicyValidationResponse>(stdout.as_bytes()) {
-                    Ok(pvr) => AdmissionResponse::from_policy_validation_response(
-                        request.uid().to_string(),
-                        request.0.get("object"),
-                        &pvr,
-                    )
+                    Ok(pvr) => {
+                        let req_json_value = serde_json::to_value(request).unwrap();
```
flavio, line 123: Can you avoid doing the `unwrap` here? We should return a rejection because of an internal error

flavio (changes requested):
Everything looks good to me, I left some minor suggestions.

I think it's fine to merge this code into the `main` branch once the review is green. I don't consider this a POC anymore, the code is actually working and delivering the whole feature

src/runtimes/wasi_cli/runtime.rs
fabriziosestito: this error can never happen since the request is serializable

flavio: > This conversion can fail if T’s implementation of Serialize decides to fail, or if T contains a map with non-string keys.

We can rule out the 2nd case, but we cannot assume the serialization of the request will always be successful. I think there are corner cases like memory allocation errors and similar. Am I too paranoid?

fabriziosestito: agreed on `except` in Slack

flavio (approved):
thanks for having applied the requested changes

PR #368, opened 2023-10-25
feat: add policy type to metadata

fabriziosestito:
## Description
Adds `policyType` (kubernetes/raw, defaults to kubernetes) to the Policy metadata.

## Test

yes

### DIFF
--- a/src/policy_artifacthub.rs
+++ b/src/policy_artifacthub.rs
@@ -474,7 +474,7 @@ fn parse_annotations(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::policy_metadata::ContextAwareResource;
+    use crate::policy_metadata::{ContextAwareResource, PolicyType};
     use assert_json_diff::assert_json_eq;
     use serde_json::json;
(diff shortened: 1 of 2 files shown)


viccuad (approved):
LGTM. I would prefer to explicitly add `policyType: kubernetes` to the `metadata_backwards_compatibility` line 385.

src/policy_artifacthub.rs
```diff
@@ -509,6 +509,7 @@ mod tests {
             background_audit: true,
             context_aware_resources: HashSet::new(),
             execution_mode: Default::default(),
+            policy_type: Default::default(),
```
flavio, line 512: I think we should be explicit and make it `PolcyType::Kubernetes`

flavio (commented):
LGTM, but I agree with @viccuad's comment

src/policy_artifacthub.rs
fabriziosestito: done in https://github.com/kubewarden/policy-evaluator/pull/368/commits/8086300c601b4b29b1d5cac15213381b17bb88eb

Issue #378, opened 2023-11-13
tech debt: extend test suite to evaluate policies


flavio:
Currently the test suite of policy-evaluator is not evaluating a policy. These tests are currently done inside of kwctl/policy-server e2e tests.
That means that, when breaking policy-evaluator, the bugs are noticed only later on inside of kwctl or policy-server.

It would be great to perform some real evaluations also inside of the tests of policy-evaluator, checking that the runtimes are working fine.

Note: I would prefer to NOT have to commit wasm modules to the git repository. I think we can leverage policy-fetcher to download the test policies and then evaluate them

fabriziosestito:
Tests to add:
- [ ] Evaluate a OPA policy (simple, context-aware)
- [ ] Evaluate a Gatekeeper policy (simple, context-aware)
- [ ] Evaluate a WAPC policy (simple, context-aware)
- [ ] Evaluate a WASI policy

flavio:
this is now done, can be closed

PR #389, opened 2023-11-23
refactor: Make Rego inventories deterministically ordered with BTreeMap

viccuad:
## Description

refactor: Make Rego inventories ordered with BTreeMap. This increases single access complexity, but we normally read all values.

This is to be consumed by kwctl for deterministic Rego context-aware inventories.

## Test

CI

### DIFF
--- a/src/policy.rs
+++ b/src/policy.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use std::clone::Clone;
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 use std::fmt;
 use tokio::sync::mpsc;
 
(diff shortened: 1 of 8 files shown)


viccuad:
We decided to not add tests in policy-evaluator for the deterministic order of `ContextAwareResources` as it needs to be done in integration tests, so we postponed to https://github.com/kubewarden/policy-evaluator/issues/378#issuecomment-1824620206. Unit tests are not useful; it is not enough to check against fixtures, or that the structures implement `Ord`.

flavio (approved):
LGTM, thanks for the fix

PR #393, opened 2023-11-27
refactor: remove wapc EvaluationContext global state and locking

fabriziosestito:
## Description

This PR removes the need to use a global state to store the retrieve the `EvaluationContext` of a waPC policy when the policy is using a host callback.
It introduces the `new_host_callback` constructor that takes a reference (Arc/Mutex) to the evaluation context and returns a higher-order function (the callback).
This removes the need for locking several times since the host callback function does not need to retrieve the `EvaluationContext` from a global state anymore.

## Test
Existing tests were updated and refactored.
Lock-related tests were removed.
Also, the `policy_evaluator_builder` module was refactored as `policy_evaluator` sub-module.
This simplifies the testing, since this way, the tests in the `policy_evaluator_builder` can access private fields of the parent module.


flavio (approved):
Looks great, not only we managed to remove the locks, but we also simplified the code base a lot!

flavio:
I ran the load tests written with k6 against policy-server. The performance of this branch if comparable with the latest rewrite I introduced; the numbers are basically the same.

However, the policies being tested are not generating any context aware request. I a load test made against this PR + policies making context aware requests (like `log`statements) will outperform the code we currently have. I know @fabriziosestito is writing a policy to be used for this testing scenario

PR #396, opened 2023-12-06
feat: introduce `PolicyEvaluatorPre`

flavio:
Introduce pre-initialized stack for all the types of policies that we currently support (Rego, waPC and WASI). These pre-initialized instance are managed by a high level resource called `PolicyEvaluatorPre`.

`PolicyEvaluatorPre` is created by `PolicyEvaluatorBuilder` once, then it can be used to create as many `PolicyEvaluator` instances as needed.

The `PolicyEvaluatorPre` instance is an immutable structure that can be safely shared across different threads once wrapped into a `Arc`.

This change is required to further reduce the memory usage of Policy Server, without compromising our performance.

Note: this PR is currently consuming a fork of waPC. I'm waiting for upstream to merge my PR.

### DIFF
--- a/src/policy_evaluator/policy_evaluator_builder.rs
+++ b/src/policy_evaluator/policy_evaluator_builder.rs
@@ -1,14 +1,12 @@
 use anyhow::{anyhow, Result};
-use std::collections::BTreeSet;
 use std::path::Path;
-use tokio::sync::mpsc;
 use wasmtime_provider::wasmtime;
 
-use crate::callback_requests::CallbackRequest;
 use crate::evaluation_context::EvaluationContext;
-use crate::policy_evaluator::{PolicyEvaluator, PolicyExecutionMode};
-use crate::policy_metadata::ContextAwareResource;
-use crate::runtimes::{rego::BurregoStack, wapc::WapcStack, wasi_cli, Runtime};
+use crate::policy_evaluator::{
+    policy_evaluator_pre::StackPre, PolicyEvaluator, PolicyEvaluatorPre, PolicyExecutionMode,
+};
+use crate::runtimes::{rego, wapc, wasi_cli, Runtime};
 
 /// Configure behavior of wasmtime [epoch-based interruptions](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.epoch_interruption)
 ///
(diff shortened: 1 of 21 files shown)


src/policy_evaluator/policy_evaluator_builder.rs
```diff
@@ -203,12 +176,88 @@ impl PolicyEvaluatorBuilder {
...
+                );
+                StackPre::from(rego_stack_pre)
+            }
+        };
+
+        Ok(PolicyEvaluatorPre { stack_pre })
+    }
+
     /// Create the instance of `PolicyEvaluator` to be used
+    // TODO: later on we can drop this method and force also kwctl to use `build_pre`
```
fabriziosestito, line 223: shouldn't we drop this in this PR?

fabriziosestito (approved):
LGTM!

The only doubt I have is if we want to keep the `build` method around.

src/runtimes/rego/stack_pre.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub entrypoint_id: i32,
+    pub policy_execution_mode: RegoPolicyExecutionMode,
+}
+
+impl StackPre {
+    pub(crate) fn new(
+        engine: wasmtime::Engine,
+        module: wasmtime::Module,
+        epoch_deadlines: Option<EpochDeadlines>,
+        entrypoint_id: i32,
```
viccuad, line 31: Nit: If we are using always the same entrypoint `0` for now, maybe we can set it inside new() and not accept it as an argument.

viccuad (approved):
LGTM! amazing!

src/policy_evaluator/policy_evaluator_builder.rs
flavio: Done via 8752bbda6a1599a7ed877735bed64fb50ff0c643

src/runtimes/rego/stack_pre.rs
flavio: I prefer to leave the arguments around, because if you put a non-zero value in there everything will work as expected. The only place where the entrypoint is hard coded is the builder.

fabriziosestito (approved):
:shipit:

Issue #377, opened 2023-11-13
tech debt: implement custom error types


flavio:
Currently the crate makes use of `anyhow::Result` to handle errors. We should implement crate-specific errors using `thiserror`, like we already do with other projects we have.

## Acceptance criteria

- [ ] All public errors of the crate are typed  using `thiserror`.

After evaluation decided that the following has diminished returns and will not be tackled:
-  Errors used for communicating in the host callbacks are typed using `thiserror`.

PR #424, opened 2024-01-29
refactor: Partially drop anyhow from `crate::runtimes:wapc`

viccuad:
## Description

Relates to https://github.com/kubewarden/policy-evaluator/issues/377.

- Partially implement thiserror a type and its enums for `crate::runtimes::wapc`.
  It is still missing `crate::runtimes::wapc::callback` which will follow up later. The definition of the callback function [type is more complicated](https://github.com/kubewarden/policy-evaluator/blob/main/src/runtimes/wapc/callback.rs#L16), and I wanted to have a full codepath for the errors first.
- Finish migration to thiserror for `crate::runtimes::rego`. 

I recommend reviewing per commit, yet note that for `protocol_version()` there's a stopgap change between commits.

Note: `protocol_version()` is only consumed by kwctl but not yet by policy-server, opened https://github.com/kubewarden/policy-server/issues/642.

## Test

CI, integration tests.

### DIFF
--- a/src/runtimes/rego/stack_pre.rs
+++ b/src/runtimes/rego/stack_pre.rs
@@ -1,7 +1,6 @@
-use anyhow::Result;
-
 use crate::policy_evaluator::RegoPolicyExecutionMode;
 use crate::policy_evaluator_builder::EpochDeadlines;
+use crate::runtimes::rego::errors::{RegoRuntimeError, Result};
 
 /// This struct allows to follow the `StackPre -> Stack`
 /// "pattern" also for Rego policies.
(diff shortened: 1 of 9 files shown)


fabriziosestito (approved):

src/runtimes/rego/stack_pre.rs
```diff
@@ -1,8 +1,8 @@
-use anyhow::Result;
-
 use crate::policy_evaluator::RegoPolicyExecutionMode;
 use crate::policy_evaluator_builder::EpochDeadlines;
 
+use super::errors::{RegoRuntimeError, Result};
```
fabriziosestito, line 4: nit: use crate and avoid super as per code conventions

viccuad: true! done inhttps://github.com/kubewarden/policy-evaluator/pull/424/commits/220620b9234e2a7016b2dd7bfd4d11cd5be0f8b4

flavio (approved):

PR #426, opened 2024-01-30
refactor: Drop anyhow from wasi_cli

viccuad:
## Description

Relates to https://github.com/kubewarden/policy-evaluator/issues/377

Opening this as draft for easy discussion.

Currently, it fails as it is expected that WasiRuntimeError can be converted from `std::result::Result<std::convert::Infallible, wasmtime::Error>`.

 

## Test

CI

### DIFF
--- a/src/runtimes/wasi_cli/errors.rs
+++ b/src/runtimes/wasi_cli/errors.rs
@@ -4,9 +4,6 @@ pub type Result<T> = std::result::Result<T, WasiRuntimeError>;
 
 #[derive(Error, Debug)]
 pub enum WasiRuntimeError {
-    #[error("cannot set wasi args")]
-    WasiStringArray(#[source] wasi_common::StringArrayError),
-
     #[error("program exited with code {code:?}; stderr set to '{stderr}', error: '{error}'")]
     WasiEvaluation {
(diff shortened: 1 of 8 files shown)


src/runtimes/wasi_cli/errors.rs
```diff
@@ -1,23 +1,55 @@
...
+    WasmLinkerError(#[source] wasmtime::Error),
+
+    #[error("cannot add to linker: {0}")]
+    WasmInstantiate(#[source] wasmtime::Error),
+
+    #[error("cannot find 'mem' export")]
+    WasiMemExport(),
+
+    #[error("'mem' export cannot be converted into a Memory instance")]
+    WasiMemExportCannotConvert(),
```
fabriziosestito, line 34: the parentheses here indicate that the error is an empty tuple variant, we can remove this and use `WasiMemExportCannotConvert` directly.
Could you change the other error variants that slipped from the past PRs?

viccuad: Doh true, done in https://github.com/kubewarden/policy-evaluator/pull/426/commits/8b41351bfb794c4e5589fd5dcd08e9165dd4c712

Issue #428, opened 2024-01-29
Implement the OCI image index capability


jvanz:
Implement inside the policy-evaluator the new capability used to fetch OCI image index manifest. Which can be used to check the platforms available for a given OCI image.

### Acceptance criteria
- Updates policy-evaluator adding the new capability function in the [`Client`](https://github.com/kubewarden/policy-evaluator/blob/275e402a2f02655babcb68e4a99ad06bd9891c9c/src/callback_handler/oci.rs#L13) to fetch the image manifest
- Add tests for the new capability

PR #429, opened 2024-01-30
feat: image manifest capability.

jvanz:
THIS PR IS WAITING FOR THE https://github.com/kubewarden/docs/pull/347 TO GET MERGE BEFORE MARKING IT AS READY FOR REVIEW. THIS IS JUST A INITIAL CODE. 
## Description

Adds a new OCI capability to allow users to request the OCI image manifest.

Fix https://github.com/kubewarden/policy-evaluator/issues/428


src/callback_requests.rs
```diff
@@ -34,6 +34,10 @@ pub enum CallbackRequestType {
         /// String pointing to the object (e.g.: `registry.testing.lan/busybox:1.0.0`)
         image: String,
     },
+    OciManifest {
```
viccuad, line 37: I would like to also have a comment explaning what does this callback request entail, like the others.

src/runtimes/callback.rs
```diff
@@ -102,6 +102,29 @@ pub(crate) fn host_callback(
                         eval_ctx,
                     )
                 }
+                "v1/image_manifest" => {
```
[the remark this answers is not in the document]
viccuad: Agree, I think it should be `v1/oci_manifest`. This means amending the docs too.

src/callback_handler/oci.rs
```diff
@@ -30,6 +31,17 @@ impl Client {
         serde_json::to_string(&image_digest)
             .map_err(|e| anyhow!("Cannot serialize response to json: {}", e))
     }
+
+    pub async fn manifest(&self, image: &str) -> Result<OciManifest> {
```
viccuad, line 35: is it ok to return the OciManifest type? shouldn't we be returning a String with the serialized json?. I suppose it happens automatically anywas once we send the callback  back. But then I wonder why for `digest()` we do the conversion to String.

flavio: Yeah, I think it would be better to serialize that to JSON and then just return the JSON string

jvanz: I'm not sure if this is necessary. The digest is a string because the digest will be store inside the `ManifestDigestResponse` type (see `get_oci_digest_cached`) wich will be parsed into a [JSON string later](https://github.com/kubewarden/policy-evaluator/blob/26c4ec2143542c10293dab0c0c0b11716fe51a2c/src/callback_handler.rs#L45) on. Therefore, I do not see the reason of serializing the content here. Furthermore, if we return string, the response will be a string where the content is a serialized JSON. And not a serialized JSON value. Because we serialized it twice. Which means that the other end will need to serialize twice as well.  Once to get the string and the other to transform the string into a JSON value. 

Does it make sense? Am I missing something?

jvanz: What have I noticed is that the function is using a type from oci-distribution. And the rust SDK does not have it as a dependency. Because it does not import policy-fetcher and policy-evaluator. Therefore, we need to add this edge in  dependency graph (maybe using a transitive dependency). I remember that we change the dependency orientation some time ago for some reason. I guess is to simplify the dependency bumps. Another option is doing something similar to what the digest does and create a type in the SDk and use it here as well.

flavio: Ok, scratch that. The host is going to serialize the `OciManifest` into JSON, and return that to the guest.

In terms of the SDKs, I would say it's up to us to decide how to handle that. The Rust SDK can provide a `get_manifest` a function that:

* Calls the host function
* Handles the error
* deserializes the JSON object into a Rust type
* return the Rust type

I think we should avoid to use the oci-distribution crate, since it's too fat for what we need. We should instead use the [`oci-spec`](https://crates.io/crates/oci-spec/) one, which contains only the definitions of the objects. In the beginning I would just introduce this extra dependency. The rust compiler should be smart enough and not add the library to the final `.wasm` file unless the code is actually used.

As for the Go one, we could do the very same OR we could just have the SDK return a generic [`RawMessage`](https://pkg.go.dev/encoding/json#RawMessage).
I would prefer to have a well defined Go Type to be returned, but we will have to find something that 1) builds with TinyGo 2) doesn't make all the policies bigger (especially the ones that are not using this capability)

These are however topics that need to be discussed when looking at the actual Rust/Go SDK.

This is a small feature, but it involves different projects. The way we tackled that in the past was to make PRs using code from PR branches. Go ahead and build a Rust policy that makes use of this capability, using a Rust SDK from a feature branch and a kwctl that consumes policy-evaluator from this PR branch

jvanz: I've opened an [PR](https://github.com/kubewarden/policy-sdk-rust/pull/129) adding the capability in the Rust SDK using the `oci-spec`. Take a look on it as well.

jvanz: @flavio @viccuad,  can we consider this PR fine? As far as I can see we can keep things as they are. Right?

flavio: Yes, I think this is fine. However let's not merge it yet, at least not until we have tested kwctl with a policy written in Go and Rust that make use of this new capability

jvanz:
@kubewarden/kubewarden-developers, [I think ](https://github.com/kubewarden/policy-server/issues/638#issuecomment-1924482913)we are good to go and start merging PR. But we need to release `policy-fetcher` before merging this PR. We need to get the `oci-distribution` version `0.10.0` to avoid a bug fixed on this [commit](https://github.com/krustlet/oci-distribution/commit/0036d93e8f6eb1df4e01e60c4f1c9c77ea43461c).

tests/integration_test.rs
```diff
@@ -298,3 +304,59 @@ async fn test_runtime_context_aware<F, Fut>(
         .send(())
         .expect("cannot send shutdown signal");
 }
+
+#[rstest]
+#[case::policy("ghcr.io/kubewarden/tests/context-aware-test-policy:latest")]
```
fabriziosestito, line 309: I think we need to use a brand new test policy here since the context-aware one needs the Kubernetes mock scenario to run.
Maybe we could have a host capabilities test policy with to test all the other callbacks including the image manifest one.

tests/integration_test.rs
```diff
@@ -214,7 +220,7 @@ async fn test_policy_evaluator(
 )]
 #[case::wapc(
     PolicyExecutionMode::KubewardenWapc,
-    "ghcr.io/kubewarden/tests/context-aware-test-policy:v0.1.0",
+    "ghcr.io/jvanz/policies/context-aware-test-policy:latest",
```
fabriziosestito, line 223: see comment: https://github.com/kubewarden/policy-evaluator/pull/429/commits/ee4a2d7d857472024a4b40cac2f4ce069cb28456#r1479439169

tests/integration_test.rs
viccuad: Agree, I suppose the changes in `ghcr.io/jvanz/policies/context-aware-test-policy:latest` will end up in `ghcr.io/kubewarden/tests/context-aware-test-policy`.

viccuad (approved):
LGTM!
From what I understand, once the accompanying context-aware example policy for the integration tests is present it can be merged.

tests/integration_test.rs
jvanz: This change is wrong. It's left over of other test. I'll remove it

tests/integration_test.rs
jvanz: Yes, I've started to change a policy to test this capability. But then I figured out that I can test this specific capability in isolation by spawning the callback handler alone and simulating the request sent by the evaluator. Therefore, I've cancel the policy change. Therefore, this test is not  testing the whole evaluation context it does check the OCI client integration only. Furthermore, the logic of sending request using the whole evaluation context is under test in the context aware test in the same file. Thus, I think we can create the new policy when we start to write test to test all the capabilities in the near future (considering our discussions about improving testing )

jvanz: I haven't check the content to avoid failing tests if the response from the registry (ghcr.io) change for some reason outside of out control. Furthermore, if the payload changed the serialization will fail any way. But I see the value of checking if the enum type, in other words, the content type is the expected considering the given image URI.

jvanz: Done. I've added the code to check the result type.

jvanz:
@kubewarden/kubewarden-developers talked about this PR during one of our calls. We are all fine. I'll merge it.

PR #432, opened 2024-02-06
feat!: Provide typed errors for public functions

viccuad:
## Description

Fix https://github.com/kubewarden/policy-evaluator/issues/377

Provide new typed errors: `MetadataError`, `ResponseError`, `PolicyEvaluatorPreError`, `PolicyEvaluatorBuilderError`, and return them in their respective public functions.

## Test

CI

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -54,3 +54,9 @@ pub enum MetadataError {
         error: serde_json::Error,
     },
 }
+
+#[derive(Error, Debug)]
+pub enum ResponseError {
+    #[error("cannot deserialize JSONPatch: {0}")]
+    Deserialize(#[source] serde_json::Error),
+}
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -40,7 +40,32 @@ pub enum PolicyEvaluatorError {
 }
 
 #[derive(Error, Debug)]
-pub enum PolicyEvaluatorBuilderPreError {
+pub enum PolicyEvaluatorBuilderError {
+    #[error("cannot convert given path to String")]
+    ConvertPath,
+
+    #[error("protocol_version is only applicable to a Kubewarden policy")]
+    InvokeWapcProtocolVersion(#[source] crate::runtimes::wapc::errors::WapcRuntimeError),
+
+    #[error("configuration of wasmtime is incorrect: {0}")]
+    InvalidUserInput(#[source] anyhow::Error),
+
+    #[error("error when creating wasmtime engine: {0}")]
+    WasmtimeEngineBuild(#[source] wasmtime::Error),
+
+    #[error("error when building wasm module: {0}")]
+    WasmModuleBuild(#[source] wasmtime::Error),
+
+    #[error("error when building wapc precompiled stack: {0}")]
+    NewWapcStackPre(#[source] crate::runtimes::wapc::errors::WapcRuntimeError),
+
+    #[error("error when building wasi precompiled stack: {0}")]
+    NewWasiStackPre(#[source] crate::runtimes::wasi_cli::errors::WasiRuntimeError),
+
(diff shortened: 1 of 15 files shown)


src/policy_evaluator/policy_evaluator_builder.rs
```diff
@@ -126,20 +132,20 @@ impl PolicyEvaluatorBuilder {
     }
 
     /// Ensure the configuration provided to the build is correct
-    fn validate_user_input(&self) -> Result<()> {
+    fn validate_user_input(&self) -> anyhow::Result<()> {
```
jvanz, line 135: Why are we still using anyhow here? Is to avoid rewriting all the errors? I'm aware that this errors are internal to the library. But I would like to know if there is some other reason to keep the anyhow reference here. Could not we replace this anyhow error by a custom error where it has different error messages? Maybe the `InvalidUserInput` which is the error that we map these one to. 

I do not consider this suggestion a blocker to merge this PR.

jvanz (approved):
LGTM. But I have a question about an internal function.

src/policy_evaluator/policy_evaluator_builder.rs
viccuad: indeed, it's because they are internal. It maybe be better to provide another class of errors and do an `#[error("transparent")]`. Will have a look into it.

viccuad: done in https://github.com/kubewarden/policy-evaluator/pull/432/commits/3dac38dacd18f0b45121c3bb820954ac5de60855.

src/errors.rs
```diff
@@ -38,3 +40,57 @@ pub enum PolicyEvaluatorError {
...
+    InvokeWapcProtocolVersion(#[source] crate::runtimes::wapc::errors::WapcRuntimeError),
+
+    #[error(transparent)]
+    InvalidUserInput(#[from] InvalidUserInputError),
+
+    #[error("error when creating wasmtime engine: {0}")]
+    WasmtimeEngineBuild(#[source] anyhow::Error),
+
+    #[error("error when building wasm module: {0}")]
+    WasmModuleBuild(#[source] anyhow::Error),
```
flavio, line 59: These should be converted from `anyhow::Error` to `wasmtime::Error`

src/errors.rs
```diff
@@ -38,3 +38,57 @@ pub enum PolicyEvaluatorError {
...
+
+#[derive(Error, Debug)]
+pub enum MetadataError {
+    #[error("cannot read metadata from path: {0}")]
+    Path(#[source] std::io::Error),
+
+    #[error("cannot parse custom section of wasm module: {0}")]
+    WasmPayload(#[source] wasmparser::BinaryReaderError),
+
+    #[error("cannot deserialize custom section `io.kubewarden.metadata` of wasm module: {0}")]
```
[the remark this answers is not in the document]
flavio: I would make this error more generic, like:

```rust
    #[error("cannot deserialize custom section `{section}` of wasm module: {error}")]
    Deserialize { section: String, error: String },
```

Or we can make a super specific one:

```
    #[error("cannot deserialize kubewarden metadata section: {0}")]
    Deserialize(#[source] serde_json::Error),
```

Probably the second would be better

flavio (changes requested):
it looks good, I left some minor change requests

src/errors.rs
viccuad: true! done and rebased.

src/errors.rs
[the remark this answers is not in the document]
viccuad: True! went with the first approach.

flavio (approved):

fabriziosestito (approved):

PR #438, opened 2024-02-14
fix: use informer pattern to keep k8s resources in sync

flavio:
Change the approach used to retrieve list of resources from Kubernetes. Instead of fetching all the resources and keeping them in memory for 60 seconds, move to what kubernetes calls 'informer pattern'.

The new code fetches all the requested resources like before, but then it uses a Kubernetes watch to keep them in sync.

This works by using the `reflector` functionality offered by `kube-rs`.

Fixes https://github.com/kubewarden/kubewarden-controller/issues/575


src/callback_handler/kubernetes/client.rs
```diff
@@ -0,0 +1,241 @@
...
+                .split_once('/')
+                .ok_or_else(|| anyhow!("cannot determine group and version for {api_version}"))?,
+        };
+
+        let kube_resource = KubeResource {
+            resource: kube::api::ApiResource {
+                group: group.to_string(),
+                version: version.to_string(),
+                api_version: api_version.to_string(),
+                kind: kind.to_string(),
```
jvanz, line 75:
```suggestion
                group: group.to_owned(),
                version: version.to_owned(),
                api_version: api_version.to_owned(),
                kind: kind.to_owned(),
```



[Use `to_owned()`. ](https://github.com/kubewarden/.github/blob/main/CONTRIBUTING.md#rust-code-conventions)

src/callback_handler/kubernetes/client.rs
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    /// Build a KubeResource using the apiVersion and Kind "coordinates" provided.
+    /// The result is then cached locally to avoid further interactions with
+    /// the Kubernetes API Server
+    async fn build_kube_resource(&mut self, api_version: &str, kind: &str) -> Result<KubeResource> {
+        let avk = ApiVersionKind {
+            api_version: api_version.to_string(),
+            kind: kind.to_string(),
```
jvanz, line 29:
```suggestion
            api_version: api_version.to_owned(),
            kind: kind.to_owned(),
```

jvanz (changes requested):

flavio:
I'm going to look into the code coverage failures

flavio:
Everything is green now

src/callback_handler/kubernetes/reflector.rs
```diff
@@ -0,0 +1,133 @@
...
+        let stream = watcher(api, filter).map_ok(|ev| {
+            ev.modify(|obj| {
+                // clear managed fields to reduce memory usage
+                obj.managed_fields_mut().clear();
+            })
+        });
+
+        let rf = kube::runtime::reflector(writer, stream);
+
+        let _join_handle = tokio::spawn(async move {
```
fabriziosestito, line 100: is the blackhole assignment needed here?

fabriziosestito (approved):

src/callback_handler/kubernetes/reflector.rs
flavio: we don't actually need it, I've dropped it

flavio:
@jvanz can I go ahead and merge it?

flavio:
merged as discussed during the daily

PR #456, opened 2024-03-07
fix: reduce CPU required by policies with host capabilities

flavio:
Reduce the amount of CPU used by policies with host capabilities. This is will reduce the latency of evaluating certain types of policies.

Part of the efforts to fix https://github.com/kubewarden/policy-server/issues/692


jvanz (changes requested):
The removal of the busy wait looks fine to. But we need to take a look on why the integration tests are failing

flavio:
@jvanz I've fixed the integration tests. They should all be green now

PR #465, opened 2024-03-25
gatekeeper inventory optimization

flavio:
Cache the serialized version of a Gatekeeper policy for later reuse. This is going to reduce the evaluation time of Gatekeeper policy that make use of context aware data.

Fixes https://github.com/kubewarden/policy-server/issues/704

### DIFF
--- a/crates/burrego/src/errors.rs
+++ b/crates/burrego/src/errors.rs
@@ -16,6 +16,7 @@ pub enum BurregoError {
     #[error("{msg}: {source}")]
     JSONError {
         msg: String,
+        #[source]
         source: serde_json::Error,
     },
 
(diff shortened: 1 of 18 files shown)


crates/burrego/src/errors.rs
```diff
@@ -13,8 +13,11 @@ pub enum BurregoError {
     #[error("Rego wasm error: {0}")]
     RegoWasmError(String),
 
-    #[error("JSON error: {0}")]
-    JSONError(String),
+    #[error("{msg}: {source}")]
+    JSONError {
+        msg: String,
+        source: serde_json::Error,
```
viccuad, line 19:
```suggestion
        #[source]
        source: serde_json::Error,
```

src/callback_requests.rs
```diff
@@ -177,6 +176,22 @@ pub enum CallbackRequestType {
         /// Singular PascalCase name of the resource
         kind: String,
     },
+
+    /// Is the data of the reflector tracking this query changed since the given instant?
```
viccuad, line 180: Quibble: this could be reworded to not look like a TODO question.

fabriziosestito (approved):

PR #490, opened 2024-04-30
fix: reflector initial list error handling

### LINKED ISSUE #489: Callback k8s client hangs when a reflector cannot be initialized
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The first time a request using a reflector is received, we initialize the reflector for the given resource/selectors. 
If the fetching of the initial resource list fails, the reflector does not return as ready, therefore the client hangs indefinitely. As a result, the policy waits till the Kubernetes API server timeout is reached.

### Expected Behavior

If the reflector initialization fails, the client should return an error.

fabriziosestito:
## Description

This is a "backport" of a fix that will be included as a part of:  https://github.com/kubewarden/policy-server/issues/716
Fixes #489 

## Tests
Improves wapc/wasi integration tests by adding the right watch mock API response.
Also, it adds a scenario to test the reflector initialization error.

## Note

In this implementation we are just circuit-breaking the stream, this means a generic error will be returned to the client.
This is a temporary trade-off since we are still using the default `Writer` from `kube-rs` which has its ready channel.
In `1.13` we will implement our custom writer, and the ready channel will be extended to return a result, therefore the user will receive a custom error.

### DIFF
--- a/src/callback_handler/kubernetes/reflector.rs
+++ b/src/callback_handler/kubernetes/reflector.rs
@@ -7,7 +7,7 @@ use kube::{
 };
 use std::hash::Hash;
 use tokio::{sync::watch, time::Instant};
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 use crate::callback_handler::kubernetes::KubeResource;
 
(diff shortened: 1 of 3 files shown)


viccuad (approved):
LGTM

src/callback_handler/kubernetes/reflector.rs
```diff
@@ -23,12 +23,21 @@ where
...
 {
-    stream.inspect_ok(move |event| {
-        if let Err(err) = last_change_seen_at.send(Instant::now()) {
-            warn!(error = ?err, "failed to set last_change_seen_at");
-        }
-        writer.apply_watcher_event(event)
-    })
+    stream
+        .take_while(|event| {
+            if let Err(watcher::Error::InitialListFailed(err)) = event {
```
flavio, line 28: Maybe add a couple of lines explaining why we care to handle only this specific error type

flavio (approved):
LGTM, I left a minor comment that can be ignored if you want to

PR #497, opened 2024-05-08
Tag 1.17.2

jvanz:
## Description

Updates dependencies and release the fixes in the reflectors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ authors = [
 ]
 edition = "2021"
 name = "policy-evaluator"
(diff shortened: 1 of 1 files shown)


flavio:
Temporarily put on hold. We have to discuss if @fabriziosestito's fix should be reverted. Let's talk about that during one of our dailies

flavio:
Closing, we are going to partially revert the fix that @fabriziosestito wrote. The reflector bootstrap has become too slow, which can lead to the policy reaching the evaluation timeout, hence being killed, hence the request gets rejected.

PR #519, opened 2024-06-10
feat: get OCI image manifest and configuration.

### LINKED ISSUE #518: Update policy-evaluator to expose the oci-distribution method to get the container image config.
As described in the issue kubewarden/kubewarden-controller#749, we need to update policy-evaluator to expose a method from oci-distribution crate to get that the container image [config](https://github.com/krustlet/oci-distribution/blob/4d75d90e175946b78a564a49d4b89dfc83b29344/src/client.rs#L946). 

#### Acceptance criteria
- [ ] Add callback method to allow policies to get the container image config from a given image. Something similar to what is done to get the container image [manifest](https://github.com/kubewarden/policy-evaluator/blob/64dfcff1279384268fddab0b2d42c3643f57a948/src/callback_handler/oci.rs#L35)
- [ ] Add tests

jvanz:
## Description

Adds a new host capability to allow guest to request an OCI image manifest, its digest and configuration.

Fix #518 

## Test

```shell
make test
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ picky = { version = "7.0.0-rc.8", default-features = false, features = [
   "chrono_conversion",
   "x509",
 ] }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.8.6" }
+policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.8.7" }
 semver = { version = "1.0.22", features = ["serde"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -7,6 +7,7 @@ use core::panic;
 use hyper::{Request, Response};
 use kube::client::Body;
 use kube::Client;
+use policy_fetcher::oci_distribution::manifest::OciImageManifest;
 use rstest::*;
 use serde_json::json;
 use std::collections::BTreeSet;
@@ -15,7 +16,7 @@ use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 use tower_test::mock::Handle;
 
-use policy_fetcher::oci_distribution::manifest::OciManifest;
+use policy_fetcher::oci_distribution::manifest::{OciDescriptor, OciManifest};
 
 use policy_evaluator::{
     admission_request::AdmissionRequest,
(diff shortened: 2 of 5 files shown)


viccuad (approved):
LGTM, pending policy-fetcher.

Cargo.toml
```diff
@@ -36,7 +36,7 @@ picky = { version = "7.0.0-rc.8", default-features = false, features = [
   "chrono_conversion",
   "x509",
 ] }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.8.6" }
+policy-fetcher = { git = "https://github.com/jvanz/policy-fetcher", branch = "issue518" }
```
fabriziosestito, line 39: memo: change this before merging

tests/integration_test.rs
```diff
@@ -382,3 +383,129 @@ async fn test_oci_manifest_capability(
...
+        .expect("missing callback channel");
+    assert!(cb_channel.try_send(req).is_ok());
+
+    // wait for the response
+    match rx.await {
+        Ok(msg) => match msg {
+            Ok(response_raw) => {
+                let response: serde_json::Value =
+                    serde_json::from_slice(&response_raw.payload).unwrap();
+                println!("{:?}", response);
```
fabriziosestito, line 471: leftover

tests/integration_test.rs
```diff
@@ -382,3 +383,129 @@ async fn test_oci_manifest_capability(
...
+    let callback_builder = CallbackHandlerBuilder::new(callback_handler_shutdown_channel_rx);
+    let mut callback_handler = callback_builder
+        .build()
+        .await
+        .expect("cannot build callback handler");
+    let callback_handler_channel = callback_handler.sender_channel();
+
+    tokio::spawn(async move {
+        callback_handler.loop_eval().await;
+    });
```
fabriziosestito, line 444: This block of setup code seems to be a pattern in every test, maybe we could refactor it to the common module

tests/integration_test.rs
```diff
@@ -382,3 +383,129 @@ async fn test_oci_manifest_capability(
...
+                );
+            }
+            Err(e) => {
+                panic!("{}", e)
+            }
+        },
+        Err(e) => {
+            panic!("{}", e)
+        }
+    }
```
fabriziosestito, line 507: Instead of using `match` we could simply expect/unwrap twice
e.g.

```rust
    let response_raw = rx
        .await
        .expect("cannot receive response")
        .expect("cannot get response");
```
The same applies for the `test_oci_manifest_capability` test in this file, could we fix this one as well?

fabriziosestito (changes requested):

viccuad (approved):

PR #522, opened 2024-06-17
fix: WASI runtime rejection response

### LINKED ISSUE #521: WASI runtime rejection includes the stack trace in the response status message
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I've written a CEL policy that has a subtle error inside of it. I'm trying to access a list of Ingress objects returned by a context aware call without going through the `.items` attribute of the object returned by the `list()` function.

However, the message shown to the user is the following one:

```console
kubectl apply -f ingress2.yaml         
Error from server: error when creating "ingress2.yaml": admission webhook "clusterwide-unique-ingress.kubewarden.admission" denied the request: internal server error: program exited with code Some(1); stderr set to '2024/06/12 14:07:56 failed to evaluate variables: no such key: spec
', error: 'error while executing at wasm backtrace:
    0: 0x62360d - <unknown>!<wasm function 9025>
    1: 0x616d44 - <unknown>!<wasm function 8213>
    2: 0x57cc0a - <unknown>!<wasm function 5378>
    3: 0x57a879 - <unknown>!<wasm function 5348>
    4: 0x280e59 - <unknown>!<wasm function 706>
    5: 0x61f65a - <unknown>!<wasm function 8687>
    6: 0x62692f - <unknown>!<wasm function 9392>'
```

The error message, useful to debug the problem, is buried into a stack trace. The user can get confused and even scared about that.

### Expected Behavior

I would prefer the error message to be something like that:

```console
kubectl apply -f ingress2.yaml         
Error from server: error when creating "ingress2.yaml": admission webhook "clusterwide-unique-ingress.kubewarden.admission" denied the request: failed to evaluate variables: no such key: spec
```

### Steps To Reproduce

Create a policy like the following one:

```yaml
apiVersion: policies.kubewarden.io/v1
kind: ClusterAdmissionPolicy
metadata:
  name: "unique-ingress"
spec:
  module: ghcr.io/kubewarden/policies/cel-policy
  failurePolicy: Fail
  rules:
    - apiGroups: ["networking.k8s.io"]
      apiVersions: ["v1"]
      resources: ["ingresses"]
      operations: ["CREATE", "UPDATE"]
  contextAwareResources:
    - apiVersion: networking.k8s.io/v1
      kind: Ingress
  settings:
    variables:
      - name: knownIngresses
        expression: |
          kw.k8s.apiVersion("networking.k8s.io/v1").kind("Ingress").list()
      - name: knownHosts
        expression: |
          variables.knownIngresses.map(i, i.spec.rules.map(r, r.host))
      - name: desiredHosts
        expression: |
          object.spec.rules.map(r, r.host)
    validations:
      - expression: |
          1 == 1
        message: "Cannot reuse a host across multiple ingresses"
```

Then create an Ingress object:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example-ingress
  namespace: default
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: example-service
                port:
                  number: 80
```

### Environment

```markdown
- kubewarden 1.13
- v1.0.0 of the policy
```

### Anything else?

I don't know if the problem lies within this policy or if that's something affecting all the WASI policies.

fabriziosestito:
## Description

Fixes #521

### DIFF
--- a/src/runtimes/wasi_cli/errors.rs
+++ b/src/runtimes/wasi_cli/errors.rs
@@ -4,9 +4,8 @@ pub type Result<T> = std::result::Result<T, WasiRuntimeError>;
 
 #[derive(Error, Debug)]
 pub enum WasiRuntimeError {
-    #[error("program exited with code {code:?}; stderr set to '{stderr}', error: '{error}'")]
+    #[error("{stderr}")]
     WasiEvaluation {
-        code: Option<i32>,
         stderr: String,
         #[source]
         error: wasmtime::Error,
--- a/src/runtimes/wasi_cli/runtime.rs
+++ b/src/runtimes/wasi_cli/runtime.rs
@@ -72,10 +72,7 @@ impl<'a> Runtime<'a> {
                     ),
                 }
             }
-            Err(e) => AdmissionResponse::reject_internal_server_error(
-                request.uid().to_string(),
-                e.to_string(),
-            ),
+            Err(e) => AdmissionResponse::reject(request.uid().to_string(), e.to_string(), 500),
         }
     }
 
--- a/src/runtimes/wasi_cli/stack.rs
+++ b/src/runtimes/wasi_cli/stack.rs
@@ -72,7 +72,7 @@ impl Stack {
         // references to the WritePipe(s) that we need exclusive access to.
         drop(store);
 
-        let stderr = pipe_to_string("stderr", stderr_pipe)?;
+        let stderr = pipe_to_string("stderr", stderr_pipe)?.trim().to_string();
 
         if let Err(err) = evaluation_result {
             if let Some(exit_error) = err.downcast_ref::<wasi_common::I32Exit>() {
--- a/src/runtimes/wasi_cli/stack.rs
+++ b/src/runtimes/wasi_cli/stack.rs
@@ -1,5 +1,6 @@
 use std::io::Cursor;
 use std::sync::{Arc, RwLock};
+use tracing::debug;
 use wasi_common::pipe::{ReadPipe, WritePipe};
 use wasi_common::sync::WasiCtxBuilder;
 use wasi_common::WasiCtx;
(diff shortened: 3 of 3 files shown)


src/runtimes/wasi_cli/runtime.rs
```diff
@@ -72,10 +72,7 @@ impl<'a> Runtime<'a> {
                     ),
                 }
             }
-            Err(e) => AdmissionResponse::reject_internal_server_error(
-                request.uid().to_string(),
-                e.to_string(),
-            ),
+            Err(e) => AdmissionResponse::reject(request.uid().to_string(), e.to_string(), 500),
```
fabriziosestito, line 75: reject_internal_server_error adds `internal server error:`

src/runtimes/wasi_cli/errors.rs
```diff
@@ -4,9 +4,8 @@ pub type Result<T> = std::result::Result<T, WasiRuntimeError>;
 
 #[derive(Error, Debug)]
 pub enum WasiRuntimeError {
-    #[error("program exited with code {code:?}; stderr set to '{stderr}', error: '{error}'")]
+    #[error("{stderr}")]
```
fabriziosestito, line 7: removed Code as we are not using it to build the display string, however the return code is used in a debug log (see below)

flavio (approved):
Thanks, this looks better.

When I wrote the code I was naive and I bubbled up all these information without realizing they would have polluted the rejection message shown to the user

viccuad (approved):

flavio (approved):
LGTM, please go ahead and tag a new release once this PR is merged

PR #524, opened 2024-06-17
fix: missing callback for new capability.

jvanz:
## Description

The policy-evaluator is missing the mapping of the callback to the recent added capability to get container image manifest and config. This commit fix that adding the missing match arm in the callbacks.rs file.

Related to https://github.com/kubewarden/kubewarden-controller/issues/749


viccuad (approved):
LGTM. (quibble: I would have titled the commit with `feat: `, but no need to rebase as is internal lib).

flavio (approved):

PR #548, opened 2024-08-07
feat: GroupVersionKind and GroupVersionResource derive extra traits

flavio:
Derive the `Eq` and `PartialEq` traits. These are useful when writing assertions inside of unit tests.


flavio:
I would tag a patch release once this is merged. That's required by some new unit tests of kwctl

viccuad (approved):

viccuad:
shall we bump the cargo.toml version in this PR then?

flavio:
I'll handle that afterwards

PR #583, opened 2024-12-02
fix: return digest as plain string instead of a json string 

fabriziosestito:
## Description

The `OciManifestDigest` response returns the digest as a JSON string since it was erroneously encoded to JSON in the callback handler.
This PR removes the extra encoding returning a plain string.
## Test

Test was missing, added


flavio (approved):
thanks for the fix and the test!

PR #599, opened 2025-01-14
fix: tinygo 0.35.0 policies exit at startup

flavio:
The policies built with tinygo >= 0.35.0 failed to run.

The issue was inside of the wasmtime-provider of the wapc project.

This commit updates to a version of the crate that fixes the problem.


flavio:
Once merged I'll tag a new version of policy-evaluator

viccuad (approved):

PR #666, opened 2025-03-27
feat: add policy group evaluator

flavio:
Add the code required to evaluate groups of policies. This code was initially inside of Policy Server, now it's being moved it to make it reusable by kwctl.


viccuad (approved):
LGTM.

fabriziosestito (approved):

Issue #715, opened 2025-06-24
Feature Request: Host capability to check service account permissions


jvanz:
### Is your feature request related to a problem?

We want to write a policy able to check if a service account used in a workload resource has too much permissions. For this, we would like to use the Kubernetes API to check if the given service account has some not expected permissions.

### Solution you'd like

Implement a new host capability to consume the Kubernetes API to check if the given service account has some given permissions

PR #718, opened 2025-07-03
chore: bump version v0.27.0.

jvanz:
## Description

Bump the policy-evaluator version to v0.27.0 to release the new kubernetes/can-i host capability.

Related to https://github.com/kubewarden/policy-evaluator/issues/715


viccuad (changes requested):
We should consume the correct URI for the test policy:

https://github.com/kubewarden/policy-evaluator/blob/fe5b42ca511118c0c3f09cac775775234e7e6839/tests/integration_test.rs#L270-L277

jvanz:
> We should consume the correct URI for the test policy:
> 
> https://github.com/kubewarden/policy-evaluator/blob/fe5b42ca511118c0c3f09cac775775234e7e6839/tests/integration_test.rs#L270-L277

Done

PR #721, opened 2025-07-09
fix(kubernetes/can-i): remove the kubernetes object wrappper

jvanz:
## Description

Removes the struct to wrapper the Kubernetes SubjectAccessReview object and all the code allowing caching it. Now, the kubernets/can-i host capability will receive the user and resource information and build the SubjectAccessReview from them.

Related to https://github.com/kubewarden/kubewarden-controller/issues/902


tests/integration_test.rs
```diff
@@ -267,9 +267,9 @@ async fn test_policy_evaluator(
     "app_deployment.json",
     rego_scenario
 )]
-#[case::wapc_cani(
+#[case::wapc_can_i(
     PolicyExecutionMode::KubewardenWapc,
-    "ghcr.io/kubewarden/tests/context-aware-test-policy:v0.2.0",
+    "ghcr.io/jvanz/tests/context-aware-test-policy:v0.2.1",
```
jvanz, line 272: This should be updated when changes from https://github.com/kubewarden/context-aware-test-policy/pull/63 is released

src/callback_handler/kubernetes.rs
```diff
@@ -149,15 +147,19 @@ pub(crate) async fn has_list_resources_all_result_changed_since_instant(
 
 pub(crate) async fn can_i(
     client: Option<&mut Client>,
-    subject_access_review: &SubjectAccessReviewWrapper,
+    user: String,
+    group: String,
+    namespace: String,
+    resource: String,
+    verb: String,
```
flavio, line 154: I've made a suggestion on the SDK about the type of payload we should send to the host. Given the new type is more complex, I would simply pass it as an argument (I think it should be doable, I hope I'm not missing something)

src/callback_handler/kubernetes/client.rs
```diff
@@ -295,15 +297,37 @@ impl Client {
...
         &mut self,
-        subject_access_review: &SubjectAccessReview,
+        user: String,
+        group: String,
+        namespace: String,
+        resource: String,
+        verb: String,
     ) -> Result<SubjectAccessReviewStatus> {
+        let subject_access_review = SubjectAccessReview {
+            spec: SubjectAccessReviewSpec {
```
flavio, line 307: I would go into the SDK and implement the `From<SubjectAccessReview> for SubjectAccessReviewSpec` trait. This would allow us to do something like:

```rust
SubjectAccessReview {
  spec: access_review.into(),
}
```

BTW, I just realized that we named our type `SubjectAccessReview`, which would conflict/cause confusion with the same type defined by the `k8s-openapi` crate.
This is fine, I think we should leave the nice name inside of the SDK. We can just rename our own type here as soon as we import it:

```rust
use kW_sdk::SubjectAccessReview as KWSubjectAccessReview;
```

src/callback_handler/kubernetes/client.rs
```diff
@@ -295,15 +297,37 @@ impl Client {
...
             .await;
-        response
-            .map(|response| response.status.unwrap_or_default())
-            .map_err(anyhow::Error::new)
+        response.map_err(anyhow::Error::new).and_then(|response| {
+            if response.status.is_none() {
+                Err(anyhow!("SubjectAccessReview did not return a status"))
+            } else {
+                Ok(response.status.unwrap())
+            }
```
flavio, line 330: Instead of using the `if/else` you can use `Option::ok_or`

src/callback_requests.rs
```diff
@@ -209,7 +204,11 @@ pub enum CallbackRequestType {
     },
 
     KubernetesCanI {
-        subject_access_review: Box<SubjectAccessReviewWrapper>,
+        user: String,
+        group: String,
+        namespace: String,
+        resource: String,
+        verb: String,
         disable_cache: bool,
```
flavio, line 212: This will have to be updated to the new type I proposed inside of the SDK. You should be able to embed the `ResourceAttribute` straight into this enum.

BTW, please add inline docs

flavio (changes requested):

src/callback_handler/kubernetes.rs
```diff
@@ -168,14 +167,18 @@ pub(crate) async fn can_i(
...
     time = 5,
     result = true,
-    key = "String",
-    convert = r#"{ format!("can_i({})", subject_access_review) }"#,
+    // We can use the request type as key because cached requires the key to implement Hash + Eq
+    // traits. As we already implement these traits, there is no need to have a custom logic for key
+    // generation. If we do that, we will only convert it into a type (e.g. string)  that
+    // implements the traits as well. 
+    key = "KWSubjectAccessReviewRequest",
+    convert = r#"{request.clone()}"#,
```
jvanz, line 175: > For all store types, except Redis, must either be owned and implement Hash + Eq + Clone, the cached_key! macro is used with a Key block specifying key construction, or a convert expression is specified on a procedural macro to specify how to construct a key of a Hash + Eq + Clone type.

https://github.com/jaemk/cached/blob/f599223f538b2b87b52af62b6e49a3f6d6fe7482/src/stores/timed.rs#L33-L42

src/callback_requests.rs
jvanz: I've added the `SubjectAccessReviewRequest` directly on the enum. Therefore, I convert it into `SubjectAccessReview` when necessary later in the code. Before calling Kubernetes API. Take a look on it (and in the sdk trait implementations) and let me know what do you think.

flavio (approved):

src/callback_requests.rs
viccuad: Thanks, now one could test for Groups, I consider this resolved.

src/callback_handler/kubernetes/client.rs
```diff
@@ -295,15 +297,37 @@ impl Client {
...
+            spec: SubjectAccessReviewSpec {
+                user: Some(user),
+                resource_attributes: Some(ResourceAttributes {
+                    namespace: Some(namespace.to_owned()),
+                    verb: Some(verb),
+                    group: Some(group),
+                    resource: Some(resource),
+                    ..Default::default()
+                }),
+                ..Default::default()
```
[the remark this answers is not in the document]
viccuad: With the conversion to `SubjectAccessReview` we can test for both user and groups, I consider this resolved.

src/callback_requests.rs
```diff
@@ -208,8 +203,21 @@ pub enum CallbackRequestType {
...
+
+        /// Disable caching of results obtained from Kubernetes API Server
+        /// By default query results are cached for 5 seconds, that might cause
+        /// stale data to be returned.
+        /// However, making too many requests against the Kubernetes API Server
+        /// might cause issues to the cluster
+        ///
+        /// This field is also defined in the `KWSubjectAccessReviewRequest`. However, we duplicate
+        /// the value here to keep the same pattern used by the other Kubernetes requests
         disable_cache: bool,
```
viccuad, line 221: Maybe I'm missing something, but I don't see why there's a need to allow for this disable_cache. This means that we expect/send 2 `disable_cache` keys in the payload. IIUC only the one inside the KWSubjectAccessReviewRequest is honoured.

This is also contrary to what we do for GetResourceRequest too, as those payloads only have 1 `disable_cache`.

jvanz: @flavio @viccuad ,  let's decide what we want to do about this. Do you want to move the `disable_cache` to another struct and add the current `SubjectAccessReviewRequest` into it together? Something like:

```rust
struct CanIRequest {
  subject_access_review_data: SubjectAccessReviewRequest,
  disable_cache: bool
}
```

jvanz: @viccuad Sorry, I just saw your proposal [here](https://github.com/kubewarden/policy-sdk-rust/pull/187#discussion_r2200277738). Which are the same thing that I just wrote here as well. I like the idea we both came up with. I vote for it as well.

viccuad (approved):

flavio (approved):

PR #744, opened 2025-08-21
fix(deps): Remove picky dependency

### LINKED ISSUE #692: Get rid of the `picky` dependency
The `picky` dependency was used a long time ago to overcome some limitations of the `webpki` crate. This crate was also used inside of the `sigstore-rs` one, which is one of our dependencies.

However, the `webpki` limitations have been overcome by the `webpki-rustls` crate. We should get rid of `picky` and embrace `webpki-rustls`, like `sigstore-rs` did.

jvanz:
## Description

Replaces the `picky` crate with `rustls-webpki` to verify certificate fields and chains. The `rustls-webpki` crate introduces stricter checks that could break the host capability API. To maintain existing behavior, this change adds a workaround using the `x509-cert` crate to validate the `not_before` and `not_after` fields before performing the full chain validation.

This introduces a minor change in behavior. Previously, if the `not_after` field was omitted by the user, the certificate's expiration was not checked. Now, if a certificate chain is provided, `rustls-webpki` will always validate the certificate's entire validity period. It is no longer possible to validate a chain while ignoring the certificate's expiration date.

Fix #692 

## Test

```shell
make test
```

### Tradeoff

Considering stricter verification performed by the `rustls-webpki`, I've tried to not break the host capability by using other crate and other ways to validate `not_before` and `not_after`.

### Potential improvement

I think we can adopt the default verification of the `rustls-webpki` crate and skip our custom `not_after` validation. That will make the code simpler and ensure a good level verification in the certificate chain. However, I cannot remember of some policy using this host capability. Therefore, I'm not sure how impactful will be this change.


src/callback_handler/crypto.rs
```diff
@@ -82,217 +137,165 @@ impl CertificatePool {
...
+                .expect("Failed to create end entity cert");
+        let end_entity_cert = Certificate {
             encoding: CertificateEncoding::Pem,
-            data: INTERMEDIATE_CA1_PEM.as_bytes().to_vec(),
+            data: end_entity_certificate.cert.pem().as_bytes().to_vec(),
         };
+
+        // use the correct CA chain
+        let cert_chain = vec![ca_cert, root_ca_cert];
+        (end_entity_cert, cert_chain)
```
jvanz, line 263: With the adoption of the rustls-webpki to verify the certificate, it's not possible to have a CA certificate in the beginning of the chain. Therefore, this function generated a root ca, intermediate ca and a end entity (leaf) certiciate

jvanz (commented):
I cannot approve it because I'm the author of the PR. But the @flavio's changes LGTM

PR #751, opened 2025-08-26
feat: add crypto host capability in CallbackRequestType.

jvanz:
## Description

In order to allow the crypto host capability to be record and replay by kwctl is necessary to add it in the CallbackRequestType enum and call the host capability code in the Callback handler.

Fix https://github.com/kubewarden/kwctl/issues/1337


src/callback_handler/crypto.rs
```diff
@@ -16,17 +15,21 @@ struct CertificatePool {
 /// verify_certificate verifies the validity of the certificate, and if it is
 /// trusted with the provided certificate chain.
 /// If the provided certificate chain is empty, it is treated as trusted.
-pub fn verify_certificate(req: CertificateVerificationRequest) -> Result<BoolWithReason> {
+pub fn verify_certificate(
```
viccuad, line 18: Why do we need to drop the usage of the `CertificateVerificationRequest`?
I would prefer to keep it, makes book-keeping easier.

src/callback_requests.rs
```diff
@@ -221,6 +223,17 @@ pub enum CallbackRequestType {
         /// the value here to keep the same pattern used by the other Kubernetes requests
         disable_cache: bool,
     },
+
+    CryptoIsCertificateTrusted {
```
viccuad, line 227: Shouldn't we wrap `kubewarden_policy_sdk::host_capabilities::crypto_v1::CertificateVerificationRequest` here?

jvanz: I've decided to not wrapp the `CertificateVerficationRequest` in the enum to avoid the same scenario that we had with the `can-i` capability. To reduce the levels in the session file. If we wrapper the type we will get something like this:

```yaml
- type: Exchange
  request: |
    !CryptoIsCertificateTrusted
    request:
      cert:
        encoding: Pem
        data:  [...]
      cert_chain: []
      not_after: 2025-12-31T23:59:59Z
  response:
    type: Success
    payload: '{"trusted":true,"reason":""}'

```

Instead of:
```yaml
- type: Exchange
  request: |
    !CryptoIsCertificateTrusted
    cert:
      encoding: Pem
      data:[...]
    cert_chain: []
    not_after: 2025-12-31T23:59:59Z
  response:
    type: Success
    payload: '{"trusted":true,"reason":""}'

```

src/callback_handler/crypto.rs
jvanz: I would like to remove the dependency from the sdk. It looks like a inversion of dependency. Maybe have a type internally to the evaluator to group the data?

flavio: Removing `CertificateVerificationRequest` import is not going to help with that, since the `Certificate` type (which is now one of the arguments of the function) is imported from our SDK.

I don't see a problem with having `policy-evaluator` import the client SDK. We did that on purpose to reuse the same types the consumers of the crate are using (and the crate itself sends down serialized over waPC).

BTW, this function is public only inside of this crate, hence we can change its signature without worrying about the consequences for kwctl and policy-server.

src/callback_requests.rs
flavio: Please add some inline documentation explaining the purpose of this enum type (like done by the other ones).

As for what to do:

> I find that acceptable, I personally treat the session file format as opaque

Same here, I don't expect a user to write the session record file manually. I personally start by recording something and then, eventually, change it by hand using the recorded data as a starting point.

> I think that having the request typed in a struct instead of the function implementation makes it clearer to see the API, and helps with future refactors and so

That would be my gut feeling, especially the last part about reducing the chances of messing something up if we were to do a refactor of the waPC API.

Because of that, I'm also aligned with Victor. I would also keep using the `CertificateVerificationRequest` from the SDK too.

flavio (commented):
I think I side with Victor on what to do.

@jvanz: please tell us if you think the point about being safer with refactoring is wrong

viccuad (approved):
LGTM!

flavio (approved):
LGTM, we just have to fix the `Cargo.toml` import once the new version of the SDK is out

flavio:
I'm fixing the broken unit tests, sorry about that

PR #774, opened 2025-10-03
fix: allow js policies built by latest javy compiler to work

flavio:
> This is a DRAFT PR: we have to decide how we want to solve this problem, an alternative solution is the one of PR https://github.com/kubewarden/kwctl/pull/1388

Our JavaScript/TypeScript policies are written using the Javy compiler and they necessitate a custom Javy pluging to work.

The Javy plugin defines some functions used by the js VM, plus it imports the usual `host:call` function from the WebAssembly host.

Everything was working fine until the javy-plugin version `4.0.0` was released.

This release requires to build our plugin using the wasip2 target, which in turns forces the usage of the WebAssembly component model. However, at this time, the javy compiler (v7.0.0) produces wasip1 modules instead of wasm components.

Among other changes, unrelated to this commit, the intermediate usage of the component model reduces the control we have over the final name of the function that is imported from the wasm host.

As a result, the latest js policies will import the `kubewarden:javy/host:call` function instead of the traditional `host:call` one.

This commit changes our WASI runtime to define an additional host function named in the same way that javy expects.

### DIFF
--- a/src/runtimes/wasi_cli/stack_pre.rs
+++ b/src/runtimes/wasi_cli/stack_pre.rs
@@ -61,60 +61,66 @@ impl StackPre {
 }
 
 fn add_host_call_to_linker(linker: &mut wasmtime::Linker<Context>) -> Result<()> {
+    let host_call_impl = |mut caller: wasmtime::Caller<'_, Context>,
+                          bd_ptr: i32,
+                          bd_len: i32,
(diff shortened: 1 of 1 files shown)


flavio:
Converting to a PR, we decided we want this fix and the kwctl one merged

PR #775, opened 2025-10-03
feat(test-policy): add policy used to test context aware call.

### LINKED ISSUE #731: Improve Workflow for Testing New policy-evaluator Host Capabilities
When developing a new host capability in policy-evaluator (like [kubernetes/can_i](https://github.com/kubewarden/policy-evaluator/issues/715)), it's difficult to test the changes effectively. Adding end-to-end tests requires updating a testing policy, but this creates a circular dependency that prevents CI from [passing](https://github.com/kubewarden/context-aware-test-policy/pull/63).

To test a new host capability in policy-evaluator, we need to update a testing policy to use it. These testing policies live in a separate repository. The CI for the policy repository requires a version of kwctl that already contains the new policy-evaluator capability. To get that kwctl version, we need to build it with the very policy-evaluator changes we're trying to test.

This deadlock means a PR to update the testing policy can't pass its CI checks. We're forced to either [disable tests](https://github.com/kubewarden/context-aware-test-policy/pull/65) or release a pre-release version of kwctl just to get the tests to run, which isn't ideal. To break this cycle, we should move the relevant testing policies directly into the policy-evaluator Rust workspace, similar to how the burrego crate is managed.

### Acceptance criteria

- Move the [context-aware testing](https://github.com/kubewarden/context-aware-test-policy) policies used in policy-evaluator integration tests into the main Rust workspace.
- Update the [integration](https://github.com/kubewarden/policy-evaluator/blob/9fd269e240775eb4253e5e6cefef79127dcae930/tests/integration_test.rs#L272) tests to build and use these local policies instead of fetching them from a registry.

jvanz:
## Description

Updates the Rust workspace adding the policy to test wapc context aware call as a member. Therefore, it's not necessary to build and release the policy to test new host capabilities added in the policy-evaluator. Updates the integration tests to run the testing context aware policy from the Rust workspace

Fix #731 

## Test

```shell
make build test integration-tests
```


flavio (changes requested):
Overall LGTM, there are two small things to be done.

Also, what do we plan to do with the original policy? Are we going to archive the repo?

Cargo.toml
```diff
@@ -61,23 +61,26 @@ wasmtime-wasi = { workspace = true }
 webpki-roots = "1"
 
 [workspace.dependencies]
-wasi-common   = "37.0"
-wasmtime      = "37.0"
-wasmtime-wasi = "37.0"
+k8s-openapi           = { version = "0.26.0", default-features = false }
```
jvanz, line 64: It's necessary to disable the features here. Otherwise, cargo complains:

```
cargo build  --target=wasm32-wasip1 --release -p context-aware-test-policy
warning: /home/jvanz/SUSE/policy-evaluator/crates/context-aware-test-policy/Cargo.toml: `default-features` is ignored for k8s-openapi, since `default-features` was not specified for `workspace.dependencies.k8s-openapi`, this could become a hard error in the future
warning: /home/jvanz/SUSE/policy-evaluator/Cargo.toml: `default-features` is ignored for k8s-openapi, since `default-features` was not specified for `workspace.dependencies.k8s-openapi`, this could become a hard error in the future
warning: /home/jvanz/SUSE/policy-evaluator/Cargo.toml: `default-features` is ignored for k8s-openapi, since `default-features` was not specified for `workspace.dependencies.k8s-openapi`, this could become a hard error in the future
```

flavio (approved):

flavio:
> > Also, what do we plan to do with the original policy? Are we going to archive the repo?
> 
> Yes, I think we can archive the repository.

Fine with me. Let's make sure:

- We update the README of the repository to explain why archived it
- Make sure we can eventually publish an update of this test policy from this repository

flavio:
I think we need to address @viccuad's points before we archive the other policy.

> I would have preferred to import the git history of the context-aware-test-policy repo, but this is not a blocker for me.

Same for me, this is not a blocker

